Issue #13, opened 2017-09-18
UMD Conversion Status


corevo:
### Core files to be converted to [UMD](https://github.com/umdjs/umd)
- [ ] src/atoms.js
- [x] src/background.js
- [x] src/commands-api.js
- [x] src/editor.js
- [ ] src/ext-command.js
- [ ] src/formatCommand.js
- [ ] src/initial.js
- [x] src/locatorBuilders.js
- [ ] src/playback-api.js
- [x] src/prompt-injector.js
- [x] src/prompt.js
- [x] src/record-api.js
- [x] src/record.js
- [x] src/selenium-api.js
- [x] src/selenium-browserbot.js
- [x] src/targetSelector.js
- [x] src/escape.js

### The following are not planned to be converted
- src/utils.js
- src/IO/load_file.js
- src/IO/save_file.js
- src/panel/js/colResizable-1.5.min.js
- src/panel/js/command_grid.js
- src/panel/js/command_grid_toolbar.js
- src/panel/js/context_menu.js
- src/panel/js/hideshow.js
- src/panel/js/jquery-3.2.1.min.js
- src/panel/js/jquery-ui.min.js
- src/panel/js/jquery.equalHeight.js
- src/panel/js/jquery.tablesorter.min.js
- src/panel/js/jquery.tabletab.js
- src/panel/js/log.js
- src/panel/js/panelSetting.js
- src/panel/js/sortable-ui.js
- src/panel/js/testcase_grid.js

PR #15, opened 2017-09-19
Migrate record and LocatorBuilders

corevo:
Migrate the user event record engine


corevo:
I updated #13 accordingly

PR #18, opened 2017-09-24
Migrate selenium-browserbot

corevo:
Migrate the module to UMD, and convert to es2015 class definitions

### DIFF
--- a/src/commands-api.js
+++ b/src/commands-api.js
@@ -17,7 +17,7 @@
 import browser from "webextension-polyfill";
 import TargetSelector from "./targetSelector";
 import Selenium from "./selenium-api";
-const BrowserBot = window.BrowserBot;
+import BrowserBot from "./selenium-browserbot";
 const locatorBuilders = window.locatorBuilders;
 
 const selenium = new Selenium(BrowserBot.createForWindow(window));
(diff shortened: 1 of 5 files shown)


shs96c:
Be aware that this is code from the main selenium repo. It may be better to use the original source as that’ll make the process of reintegrating into the main project easier.

corevo:
Indeed, I'm still not sure how to do it, I was planning on cloning and building it?

shs96c:
Ultimately this will be in the monorepo. I’m traveling right now, but will be online for a bit on Monday and then properly on Wednesday. We can plan and plot then. :)

samitbadle:
Won't it make sense to use a new folder in the main repository, say ide3? The old IDE is very tightly coupled with Firefox and it might be a disadvantage moving forward. Is there any reason other than loss of history to not use a new folder?

shs96c:
The new code will be folded back into the main repo (probably as
`//javascript/ide`) when we're closer to an alpha launch (or perhaps
shortly after). Lots of the old code was coupled to Firefox, but there was
a sizeable chunk that wasn't, as it was based on Core, which had to work
everywhere.

PR #22, opened 2017-10-01
Compile atoms

corevo:
Load atoms using google closure, then bundle them as umd per webpack's configuration


shs96c:
Nice. I’d be trying to figure out the magic to get the atoms to compile.

corevo:
No magic, just skill 🐢

PR #25, opened 2017-10-10
Docs and links to docs on yarn.

peter-kehl:


corevo:
`yarn` is a pretty common tool in the industry nowadays, I don't think that explaining it is a re-implementation of `npm` is necessary.

Issue #39, opened 2017-11-13
Make horizontal borders draggable, the tool should be usable on low-res screens too


barancev:
Set screen resolution to 1280x720 and try to use the tool. Bottom blocks are not visible, you can't scroll to them and you can't drag borders to make them visible.

![image](https://user-images.githubusercontent.com/617090/32726898-b07c6fd0-c88b-11e7-819c-5f3183472473.png)

PR #44, opened 2017-11-19
Resize window and logs persistently

corevo:
Make the extension work on multiple resolutions, while keeping the last window size persistent through uses  
- [x] Window resizable
- [x] Persistent window size
- [x] Logs resizable


corevo:
Should fix #39 
\cc @barancev

PR #59, opened 2017-12-07
remove border when mouse leave

Jongkeun:
After clicking, if mouse leave the button, the border is removed.


corevo:
It's going to prevent the user from having tab index feedback,
Moving through the buttons using tab and shift tab will have no visual effect

PR #72, opened 2017-12-21
Add context menu

Jongkeun:
add context menu on test row and test navigation.
I will try to add menu on suite navigation.
I could not add, because It is little different with others.

and I have changed some code to remove warning.
react-cursor-position is consist of `div `tag.
`div` tag can't be inside `table` tag, so it was appeared `validateDOMNesting` warning.
This is why I have added `TestRowContainer`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "react": "^16.0.0",
     "react-autocomplete": "^1.7.2",
     "react-contenteditable": "^2.0.5",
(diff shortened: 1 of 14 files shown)


Jongkeun:
Hello @corevo , 
Sorry but could you tell me how I can fix it?

Jongkeun:
Oh It has been resolved.
Before it was not appeared error messages.

PR #73, opened 2017-12-21
Add context menu

Jongkeun:
add context menu on test row and test navigation.
I will try to add menu on suite navigation.
I could not add, because It is little different with others.

and I have changed some code to remove warning.
react-cursor-position is consist of `div` tag.
`div` tag can't be inside `table` tag, so it was appeared `validateDOMNesting` warning.
This is why I have added `TestRowContainer`.

Sorry about too many pull request....

### DIFF
--- a/src/neo/components/ContextMenuContainer/index.jsx
+++ b/src/neo/components/ContextMenuContainer/index.jsx
@@ -1,0 +1,142 @@
+import React from "react";
+import PropTypes from "prop-types";
+import "./style.css";
+import { findDOMNode } from "react-dom";
+import ReactModal from "react-modal";
+import classNames from "classnames";
+import { Transition } from "react-transition-group";
+import {observer} from "mobx-react";
+
+export const MenuDirections = {
+  Left: "left",
+  Bottom: "bottom"
+};
+
+const duration = 100;
+
+const transitionStyles = {
+  entering: {
+    opacity: 0,
+    transform: "scale(0, 0)"
+  },
+  entered: {
+    opacity: 1,
+    transform: "scale(1, 1)"
+  },
+  exiting: {
+    opacity: 0,
+    transform: "scale(0, 0)"
+  },
+  exited: {
+    opacity: 0,
+    transform: "scale(0, 0)"
+  }
+};
+
+class Menu extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this.handleClosing = this.handleClosing.bind(this);
+  }
+  static propTypes = {
+    isOpen: PropTypes.bool,
+    children: PropTypes.node,
+    node: PropTypes.any,
+    width: PropTypes.number,
+    direction: PropTypes.string,
+    padding: PropTypes.number,
+    onClick: PropTypes.func,
+    requestClose: PropTypes.func.isRequired
+  };
+  static defaultProps = {
+    isOpen: false,
+    width: 200,
+    padding: 5,
+    direction: MenuDirections.Left
+  };
+  handleClosing(e) {
+    this.props.requestClose(e);
+  }
+  render() {    
+    let directionStyles = {};
+    if(this.props.rect){      
+      if(this.props.rect.x+this.props.rect.width < this.props.rect.x + this.props.position.x + this.props.width ){
+        directionStyles = {
+          top : this.props.rect.y + this.props.position.y + "px",
+          left : this.props.rect.x + this.props.rect.width - this.props.width + "px",
+          transformOrigin: `${this.props.width / 2}px 0px 0px`
+        };
+      }else{
+        directionStyles = {
+          top : this.props.rect.y + this.props.position.y + "px",
+          left : this.props.rect.x + this.props.position.x + "px",
+          transformOrigin: `${this.props.width / 2}px 0px 0px`
+        };
+      }
+      
+    }    
+
+    return (
+      <Transition in={this.props.isOpen} timeout={duration}>
+        {(status) => (
+          <ReactModal
+            className={classNames("menu", "content")}
+            isOpen={this.props.isOpen}
+            ariaHideApp={false}
+            shouldCloseOnOverlayClick={true}
+            closeTimeoutMS={300}
+            onRequestClose={this.handleClosing}
+            style={{
+              overlay: {
+                backgroundColor: "transparent",
+                zIndex: "1000"
+              },
+              content: Object.assign({
+                width: `${this.props.width}px`
+              }, directionStyles, transitionStyles[status])
+            }}
+          >
+            <div onClick={this.props.onClick}>
+              {this.props.children}
+            </div>
+          </ReactModal>
+        )}
+      </Transition>
+    );
+  }
+}
+
+export default class ContextMenuContainer extends React.Component {  
+  constructor(props) {
+    super(props);           
+    this.close = this.close.bind(this);    
+  }
+  
+  static defaultProps = {
+    closeOnClick: true
+  };
+  
+  close(e) {    
+    e.preventDefault();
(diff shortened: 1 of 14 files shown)


Jongkeun:
Is there a way to check code indent before I try to pull request?

corevo:
Of course, you can run 'yarn lint'.
Also you don't have to resubmit a pull request, if you push to your branch it'll update the pr

Cheers,
Tom

src/neo/components/Test/index.jsx
```diff
@@ -24,6 +24,10 @@ import ListMenu, { ListMenuItem } from "../ListMenu";
 import MoreButton from "../ActionButtons/More";
 import RemoveButton from "../ActionButtons/Remove";
 import "./style.css";
+import ContextMenu from "../ContextMenu";
+import { findDOMNode } from "react-dom";
+import UiState from "../../stores/view/UiState";
```
corevo: I have deliberately avoided adding the `UiState` to the leaf components, in order to make them reusable in other parts of the application.  
Pass whatever you need down from the parent.

src/neo/components/Test/index.jsx
```diff
@@ -61,13 +69,17 @@ export default class Test extends React.Component {
...
+    index: PropTypes.number,
+    position: PropTypes.any,
+    onContextMenu: PropTypes.func
   };
   componentDidMount() {
     if (this.props.selected) {
       this.props.setSectionFocus("navigation", () => {
         this.node.focus();
       });
+      this.setState({rect: findDOMNode(this).getBoundingClientRect()}); // eslint-disable-line react/no-find-dom-node
```
corevo: This breaks the encapsulation of the menu's behavior, the context menu host knows that the menu needs to open on it and keeps in it's state it's current rect.  
Instead send a prop to the context menu in a similar fashion to how I send `opener` to menu, so that the context menu will calculate the rect, and the host will be unaware of the context menu existence and behavior.

src/neo/components/TestList/index.jsx
```diff
@@ -23,42 +23,57 @@ import classNames from "classnames";
...
 import PlaybackState from "../../stores/view/PlaybackState";
+import ReactCursorPosition from "react-cursor-position";
+
 import "./style.css";
 
 @inject("renameTest") @observer
-export default class TestList extends Component {
+export default class TestList extends Component {  
+  componentDidMount() {
+    document.addEventListener("contextmenu", this.handleContextMenu);
```
corevo: Don't mess with the document's events, it's messy, it's mutating DOM outside of react's render scope.  
Prevent the default behavior on each menu `opener` like I said, add the event to it.

src/neo/components/TestTable/index.jsx
```diff
@@ -49,52 +61,78 @@ export default class TestTable extends React.Component {
...
+              clipboard={UiState.clipboard}
+              setSectionFocus={UiState.setSectionFocus}
+              position={this.props.position}                
+            />              
+          ) : null }              
+        </tbody>
+      </table>    
+    ]);
+  }
+}
```
corevo: It's my bad for not adding this rule, but new line at end of file is required.  
I will add it to the eslint config, for the time being, please add where needed.

src/neo/stores/view/UiState.js
```diff
@@ -40,7 +40,9 @@ class UiState {
   @observable navigationHover = false;
   @observable navigationDragging = false;
   @observable pristineCommand = new Command();
-  @observable lastFocus = {};
+  @observable lastFocus = {};  
+  @observable isContextOpenEditor = [];
```
corevo: It's messy to keep the state this way, since only one menu can be open at a time, and it's state is completely internal.  
Like the menu change it's internal state to open, and close it when it requests to be closed.  
App-wise state changes are unnecessary here, since the state can be kept completely internal.

src/neo/stores/view/UiState.js
```diff
@@ -97,6 +99,13 @@ class UiState {
       } else {
         this.selectCommand(undefined);
       }
+      let i = 0;
+      for(i = 0; i <= test.commands.length; i++){
+        this.isContextOpenEditor[i]=false;
```
corevo: Due to previous comment, this is completely unnecessary, but it is a bad practice, because it prevents reordering of tests.  
If you do need to keep state of a list, keep it by `id` and not `index`

src/neo/stores/view/UiState.js
```diff
@@ -127,10 +136,10 @@ class UiState {
   }
 
   @action.bound selectCommand(command) {
-    this.selectedCommand = command;
+    this.selectedCommand = command;
```
corevo: Please remove the extra spaces, maybe your editor has an option to do it?

src/neo/stores/view/UiState.js
```diff
@@ -267,6 +276,22 @@ class UiState {
     });
     this._project.modified = false;
   }
+
+  @action.bound onContextMenuEditor(index) {
+    if((document.getElementsByClassName("ReactModal__Body--open").length) > 0 ){
```
corevo: This is an anti-pattern, it breaks React's component abstraction since you directly access the DOM,  
Use your `onContextMenu` to change an internal state like discussed above,  
Try to access the DOM only when there is no other way, and through React's mechanics, like `findDOMNode`

src/neo/components/ContextMenuContainer/index.jsx
```diff
@@ -0,0 +1,142 @@
+import React from "react";
```
corevo: Please find a way to reuse the Menu component, because this component has a very similar behavior, and will require the maintaining of two components rather than a generic single one.  
Try to do so by adding a new direction, like bottom, and left, try to add cursor, or some sort of other prop to designate that this is a context menu render.

Jongkeun:
I'll request later again.

src/neo/stores/view/UiState.js
Jongkeun: @corevo , Thanks for your comments with advice.
I haven't known much about web application because this is my first web application.😅
I want you to keep giving me advice.

I fixed about your other comments.
But I have no idea on this comment.
How can I distinguish if react-modal is opened?
I don't use array for supervising modal state anymore.
I am using 'observable' each test row for modal state.

Please tell me guide line about detecting react-modal state if you have some idea.

corevo: Keep it internal, instead of having the row explicitly set the `isOpen`, set it to true internally with the `onContextMenu` event handler, and set it back to false on `requestClose`.  
This can be done inside your own component's `state` using internal `setState`, as I've done in [MenuContainer](https://github.com/SeleniumHQ/selenium-ide/blob/master/src/neo/components/Menu/index.jsx#L131)

Jongkeun: It is fine about `ContextMenuContainer`. (`ContextMenuContainer` uses same `Menu` of `MenuContainer`.)
But It is not fine together with `MenuContainer`.

My test case is 
1. open react-modal using a `Morebutton` made by @corevo .
2. open react-modal using mouse right button.

Then, it is appeared two react-modal.
because `requestClose` was not called by mouse right button click.

This is the reason that I used `getElementsByClassName` function.
I need whether It has been opened react-modal using a `MoreButton` or not.
So if react-modal has been opened, react-modal should not to open by mouse right button.

I am worried about my horrorable english skill.
I don't know if you can understand.

Jongkeun: I found out what the problem was.
`requestClose` is called only when I switch value of `isOpen`.
I thought that is called for any events.

Then I should find the way to control state of another `Menu`.
I don't know if there is a way.
I will try to find out .

PR #76, opened 2018-01-03
Add context menu all

Jongkeun:
it still has to work on Test Suites as well. -> OK
reuse the Menu Component -> OK
to avoid adding UiState -> OK
Don't break the encapsulation and use opener ->OK
Don't use document's events -> OK
remove the extra spaces on my editor -> OK
Do not directly access the DOM -> OK

And closing it requires two clicks rather than one. -> Fail.
Because It was `Modal`. I don't know if it would be possible to change to modeless.
I tried to convert `react-modal` to `react-modelss`.
But it was not simple and different with current logic.
How about your opinion?
Shall we try to change to react-modeless?


Jongkeun:
It is something strange for building.
I will check it tomorrow.

Jongkeun:
I misunderstood about And closing it requires two clicks rather than one.
I just tested it in Ubuntu and it's not working properly.
I would fix it.

Jongkeun:
It seems like a Linux problem.
What I think was strange was that reuqestClose was not called at first click on a Linux.
I don't know the reason :) 😅

I developed and tested on Windows because I'm Windows developer.
So sorry about that.
I will keep checking about what the reason is and what the solution is.

Jongkeun:
@corevo , Could you test for me?
`React.cloneElement(this.props.opener, { key: "opener", ref: (node) => {return(this.node = node || this.node);}, onClick: this.handleClick }),` 
to
`React.cloneElement(this.props.opener, { key: "opener", ref: (node) => {return(this.node = node || this.node);},` **_onContextMenu_**`: this.handleClick }),`

Please change like this of Menu component and test.
It is same on your pc too?

corevo:
Hey, I’m on vacation, I’ll check it out next work week

Cheers,
Tom

corevo:
The issue originates by the way the events are created in the browser.  
1. `onMouseDown` targeted on the row is not handled
2. `onContextMenu` targeted on the row is prevented and opens a menu
3. `onMouseUp` targeted on the **overlay** is handled by react-modal and creates an edge case state where mouse up was called but not mouse down  
We should open an issue there and see what they think

Jongkeun:
But this sample code is fine.
[sample](https://codepen.io/jongkeun/pen/jYZNaB)
I tested this sample and I thought react-modal is normal.
How about you?

I would like to open an issue on react-modal, but I didn't understand what is cause.

Jongkeun:
It's different of react-modal's version.
I will open issue.

Jongkeun:
I opened issue on react-modal and fixed 2 days ago.
And collaborator said 
_"Thanks, @Jongkeun. This seems ok. I'll run some tests and if everything is fine, we merge this."._

Is there anything else I should fix in this pull request?

src/neo/stores/view/MenuState.js
```diff
@@ -0,0 +1,26 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { observable } from "mobx";
+
+class MenuState {
+  @observable isOpen = false;
```
corevo: Why did you find it necessary to synchronize the menu states between all of them?

corevo:
Also we'll need to wait on [this](https://github.com/reactjs/react-modal/pull/591), until it is published.

src/neo/stores/view/MenuState.js
Jongkeun: Umm... I didn't think it is necessary to synchronize.
But I don't want to open two context menu at the same time.
Just like any other menus.
**_But react-modal component is not reacted on right mouse click._**
This is the most important cause.

**_If I handle each state by array, I should reset all menu state when a menu state is changed._**
These two reason is the reason that I used state.

But I think It is fine to separate TestTable and Test.
Let me try?

corevo: I don't think it's currently an issue, as having two menus open is functionally improbable because of the overlay.  
And if the user has opened two, he can just close them, I think it might be better that menus will stay unaware of each other.

src/neo/components/Test/index.jsx
```diff
@@ -44,8 +48,13 @@ function collect(connect, monitor) {
     isDragging: monitor.isDragging()
   };
 }
-
+@observer
 export default class Test extends React.Component {
+  @observable isOpen = false;
```
corevo: I think this should be handled internally as with the normal menu.  
The Test and Row only indicate that a click has happened, not that a menu should or shouldn't open, that's up to the menu's state to decide.

Jongkeun: I missed it. 
I will remove this line.
Thank you.

src/neo/components/ContextMenu/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+import "./style.css";
+
+@observer
+export default class ContextMenu extends React.Component {
+  static propTypes = {
+    width: PropTypes.number,
+    padding: PropTypes.number,
+    onContextMenu: PropTypes.func,
+    isOpen: PropTypes.bool,
+    opener: PropTypes.any,
```
corevo: Isn't `opener` not required in case of a context menu?

src/neo/components/TestList/index.jsx
```diff
@@ -23,52 +23,61 @@ import classNames from "classnames";
 import Test, { DraggableTest } from "../Test";
 import UiState from "../../stores/view/UiState";
 import PlaybackState from "../../stores/view/PlaybackState";
+import ReactCursorPosition from "react-cursor-position";
```
corevo: I don't like this component, because it is an internal requirement of the cursor direction, can you try and calculate the position inside the direction calculation?

src/neo/stores/view/MenuState.js
Jongkeun: It means let it be?

corevo: No, I think this should be removed, as it's improbable that the user will open 2 menus.  
It is useful for a fix I plan for later, to use one `react-modal` as to not spam the DOM, then we'll need to make sure there is only one.  
But for the time being there is no need in synchronizing that state.

Jongkeun: When I was at home I misunderstood your comment. Sorry.
This state's purpose is for preventing of two modal right.
But It is not for between Test and TestRow.
I asked before at [here](https://github.com/SeleniumHQ/selenium-ide/pull/73).
image

I can't distinguish whether react-modal is opened using a `Morebutton` or closed.
So, I added this state to change whenever other modal is opened by `Morebutton`.

corevo:
Can you rebase your changes instead of merging, so that only your commits show in the diff?

Jongkeun:
I did rebase, but it had conflicts. So I did this.
Then, do I have to create new branch and cherry pick?
I don't know what is the best way.

PR #100, opened 2018-02-19
remove unnecessary console log.

Jongkeun:
Hi.
I was on honeymoon.
Is this code is necessary?


corevo:
Do you mind changing the rule to warning in the eslint config, and running it across the board?

Jongkeun:
sure, I will try tomorrow.

Jongkeun:
Thanks. :)

PR #112, opened 2018-02-26
Add context menu

Jongkeun:
There was a phenomenon that 'react-modal' couldn't be closed even with a right click.
So I opened issue(reactjs/react-modal#619) and I was replied. 
If you should not want to use `addEventListener`, I can import `react-modal` code and modify some code.

Tell me a point to modify.


Jongkeun:
I told comment because you don't like to use `addEventListener` function.
Umm... I added some `props`.
I added type of `direction` called _cursor_. This make a position of Modal to open at mouse pointer.
The reason of added `closeTimeoutMS` is if I double-click fastly on other row, same modal was reopened before other row selected. So If you didn't use this prop, default value is "300" as you would have used it.
If you want to use `ContextMenu` component, you should write `ref="contextMenu"` and below code.

> handleContextMenu(e){
>     this.refs.contextMenu.handleContextMenu(e);
>   }

I needed a way calling to `Menu`'s `handleClick` function.

Have a good travel!

packages/selenium-ide/src/neo/components/ContextMenu/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+@observer
+export default class ContextMenu extends React.Component {
+  static propTypes = {
+    width: PropTypes.number,
+    padding: PropTypes.number,
+    children: PropTypes.node,
+    direction: PropTypes.string,
+    closeTimeoutMS: PropTypes.number,
+  };
+  handleContextMenu(e) {
```
corevo: I think this should be an HOC because of 2 reasons
1. Referencing this ref directly to open the menu breaks the abstraction
2. Components need to have their own onContextMenu to call this menu to open

Think of `react-dnd` how it automatically decorates the component and adds the events on it's own, I'm thinking of something in the lines of
```jsx
<ContextMenuTarget menu={<ListMenu ...>}>
  <Suite />
</ContextMenuTarget>
```

What do you think?

Jongkeun: I tried like this, but if I will do,I thought your code was modified so much.
If you prefer this way, I try again.
Thank you for taking the example.

Jongkeun: I got it wrong totally. 
HOC is new concept to me. 
I am going to study about HOC. 
Thanks again.

Jongkeun:
I applied HOC to contextmenu. 

If you want to use this, you notice how to use on `SuitList/index.jsx`
To use `withOnContextMenu `funtion, you can get `OnContextMenu`.
Naming rule is OriginContextName`WithContextMenu`.
For example, Suite -> SuiteWithContextMenu.( It is not must. )
And you can use `SuiteWithContextMenu` like a component with `menu` parameter.

packages/selenium-ide/src/neo/components/TestList/index.jsx
```diff
@@ -23,10 +23,22 @@ import classNames from "classnames";
 import Test, { DraggableTest } from "../Test";
```
corevo: Here I export two versions of `Test` one wrapped and one not, I wrap it inside since its an internal.  
In our case no matter what we wanna wrap it, so wrap `Test` with menu, and don't render anything if it is not part of the test list view.  
Don't extract the menu from the components.

packages/selenium-ide/src/neo/components/Menu/index.jsx
```diff
@@ -160,16 +193,19 @@ export default class MenuContainer extends React.Component {
...
-      React.cloneElement(this.props.opener, { key: "opener", ref: (node) => {return(this.node = node || this.node);}, onClick: this.handleClick }),
+      (this.props.opener ?
+        React.cloneElement(this.props.opener, { key: "opener", ref: (node) => {return(this.node = node || this.node);}, onClick: this.handleClick }) : null),
       <Menu
         key="menu"
-        isOpen={this.state.isOpen}
-        node={this.node}
-        onClick={this.props.closeOnClick ? this.close : null}
-        requestClose={this.close}
+        isOpen={this.props.isContextMenu ? this.props.isOpenContextMenu : this.state.isOpen}
```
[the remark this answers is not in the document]
Jongkeun: Then how can I handle `handleClick` of  `Menu` component on another component?
Give me some hit please.
I thought of it for a month.
So I used `refs`. But this is not good as you said.

And I am going to think more about `isContextMenu`.

Thanks to wait for this PR.
You helped me learn many new things.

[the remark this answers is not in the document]
corevo: I was thinking something like
`onClick={this.prop.isContextMenu ? this.props.close || this.close : null}`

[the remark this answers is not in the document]
corevo: And then you can drop `this.props.isContextMenu`, because you just override the defaults with props this way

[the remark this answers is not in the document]
Jongkeun: I don't know what are you telling.

First, you said you don't want the menu to know that it is contained within a context menu.
I understood that I should remove `isOpenContextMenu` and handle menu's `isOpen` status by only `this.state`, isn't it? I think it is best , but it is hard to me. So I needed your help. I have no idea except for using this.props.isOpen.

Secondly, If I am going to fix `onClick` as you said, `closeOnClick` is not checked.
Then `GaugeMenu` is not working appropriately. So I left `closeOnClick`.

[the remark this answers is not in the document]
corevo: Something like this?
```jsx
<Menu
        key="menu"
        isOpen={this.props.isOpen || this.state.isOpen}
        node={this.props.eventTartget ? this.props.eventTartget : this.node}
        onClick={this.props.closeOnClick ? this.props.close || this.close : null}
        requestClose={this.props.close || this.close}
        width={this.props.width}
        direction={this.props.direction}
        padding={this.props.padding}
        position={this.props.position}
        closeTimeoutMS={this.props.closeTimeoutMS}>
        {this.props.children}
      </Menu>
```
Where `isOpen` used to be `isContextMenuOpen`

Jongkeun:
Yesterday was too late for me.
My office time is 2 AM ~ 11 AM on your country time.
I fixed something you mentioned.
Please check once again.

corevo:
Sorry for being late  
I love the changes, the `paste` change is nice too,
Just get the menus inside the components again, it just doesn't feel right, that the parent defines the menu of the child.

packages/selenium-ide/src/neo/components/Suite/index.jsx
```diff
@@ -88,14 +91,22 @@ class Suite extends React.Component {
...
   }
   render() {
+    const listMenu = <ListMenu width={130} padding={-5} opener={<MoreButton />}>
+      <ListMenuItem onClick={this.props.selectTests}>Add tests</ListMenuItem>
+      <ListMenuItem onClick={() => this.props.rename(this.props.suite.name, this.props.suite.setName)}>Rename</ListMenuItem>
+      <ListMenuItem onClick={this.props.remove}>Delete</ListMenuItem>
+    </ListMenu>;
+    //setting component of context menu.
+    this.props.setContextMenu(listMenu);
+
```
Jongkeun: I don't like this part, but I don't know the way about getting `listMenu` from child component.

corevo: Its a necessary evil, I'll try to look into it.

corevo:
I added this to v1.0.4 release notes, hopefully I'll get around to releasing it this week,  
I only need to finish #132

PR #121, opened 2018-03-07
create the slack for selenium-ide.

Jongkeun:
create the slack. 
I thought something how about create a slack only for this project.
What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,4 +55,5 @@ Here's a draft of the general tasks ahead. Feel free to pitch in and announce wh
 
 We coordinate work on the [#selenium](irc://freenode.net/selenium) IRC
 channel, which is also available on
-[Slack](https://seleniumhq.herokuapp.com).
+[Slack](https://seleniumhq.herokuapp.com).<br/>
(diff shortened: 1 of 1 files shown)


corevo:
I had talks with the rest of Selenium org, they want all talk to be coordinated centrally at SeleniumHQ slack or irc.  
But I would like a channel in the main slack account, thoughts @shs96c ?

Jongkeun:
I think to be in the main slack is better.

shs96c:
The main committers on the project are on the #selenium channel, and they use IRC (partly because Slack is a closed-source product, and we'd like an OSS project to use Free Software where possible). Any other channels other than the main one aren't mirrored.

A separate IDE channel would therefore miss input from the people most closely tied to the project.

Worse, the community works best when we all talk to each other. A separate channel fragments the community, isolating users of different parts of the Selenium project from one another and reducing the opportunity to learn from each other.

For these reasons, I'm strongly against a separate selenium-ide channel on the Slack.

Jongkeun:
I agree with you, but I felt sad about not being able to read because there was so many conversations.

PR #128, opened 2018-03-15
WIP:Suite drag drop

### LINKED ISSUE #127: In the test suite type, it is not possible to drag the positions of the included tests.
In the test suite type, it is not possible to drag the positions of the included tests.

Jongkeun:
fixed #127


corevo:
Do you wanna share your thoughts on [that](https://github.com/SeleniumHQ/selenium-ide/issues/127#issuecomment-373295881)?  
Or have the conversation here?  
I think that having setup and teardown stages in all suites will solve this.  
Relying on reordering for test execution is going to make parallelization a lot harder.

Jongkeun:
I thought it was necessary.
Now, Suite is sorted by name.
So If someone wants to change the order of the tests, he should rename the test.
He just wanna change the order of that.
As you said, if the test case has all the commands, it doesn't matter.

I thought `Suite` exists to carry out multiple tests by one click.
However, I do not think it is necessary to pay attention to make name of `Test` because of it's `Suite`.

Nevertheless, you can close it if you don't think it is necessary.

corevo:
It's not about my personal preference, but about what problem we solve and how we solve it.  
The problem is that I want to avoid having test cases that are basically setup (e.g. login/logout test cases).  
And I thought that having setup and teardown might solve that, drag and drop is just another solution, but I may miss something that I'm not aware of.

hohnan:
Sorting by name seems to be problematic in that you may need to have multiples of the same test case in order for it to fall in the correct order for the specific suite it is in. 

I am using to test sites that require entry of loan details. I like use the same test case to generate random numbers, store these as variables, and verify that they are carried out throughout the site. I will use this test case in different positions of the suite depending on the user I am creating and account for, credit needs to be reviewed vs. a straight pass. 

If we are sorting by name, I will need to put say a 3 for the first scenario and a 2 for the second because there is one less step to get to that screen.

Maybe I need to be shown a different way of thinking, idk.

corevo:
@hohnan I think tests need to have assertions, although setup stages are necessary, they aren't tests.  
It is something we need to fix, but rather than using tests as setup, I think that there is room to introduce setup.

jngale:
@corevo I am late coming to this discussion, but I agree with @hohnan. Being able to order tests other than by name is great advantage. It is a pain to have them numbered so I can control the order of execution

corevo:
@Jongkeun wanna fix the conflicts here?

Jongkeun:
I don't care.
If you allow this pr, I am going to resolve the conflicts or commit again.

corevo:
I was mistaken putting this aside, more and more people are asking for it.
If you'll resolve the conflicts I'll let that in.

Cheers,
Tom

Jongkeun:
It has a problem after I merged.
I will check what the problem is.

Jongkeun:
@corevo 
There is warning when I click `add Tests`.
I fixed the warning. 
Check it please.

corevo:
I thought last night about an issue that might occur by misusing this.
I think I have a way to keep users from misusing this, I'll share my thoughts on Monday.

Cheers,
Tom

Jongkeun:
Sure. Have a good weekend.

corevo:
I wanna talk about this, the dnd experience doesn't feel right, it's not like the table.  
Wanna ping me on slack?

Jongkeun:
Can I change this to use `DragSource` like `TestRow` ?

Jongkeun:
I think this is the simplest solution.
Do you wanna leave the drag motion of `Test`?

corevo:
Can you fix the conflicts here?, I'll get it in tomorrow.

Jongkeun:
I will throw in the towel about this PR.

PR #139, opened 2018-04-03
WIP: Eliminate gotos when migrating

corevo:
The plan is to not support the likes of `goto` in the IDE, rather introduce better alternatives like `if` and `while`, in order to do that we adding conversion of the old flow to the new one.  
<br />
Convert unstructured control flow (e.g. `goto` `gotoIf`) to structured one.  
Before merging will require the IDE to support  
 - `if`
 - `while`
 - `do`
 - `break`


mr-twistyneck:
Need any examples of Legacy scripts with Goto's in them? Happy to help!

corevo:
@mr-twistyneck that would be lovely

PR #140, opened 2018-04-04
WIP: Add control flow to the IDE

corevo:
Designing flow control for the IDE, including playback, and export.  
- [ ] break
- [ ] continue
- [ ] do endDo
- [ ] if else end
- [ ] times end
- [ ] while end

### DIFF
--- a/packages/selenium-ide/src/neo/components/TestRow/index.jsx
+++ b/packages/selenium-ide/src/neo/components/TestRow/index.jsx
@@ -106,6 +106,7 @@ class TestRow extends React.Component {
     command: PropTypes.string.isRequired,
     target: PropTypes.string,
     value: PropTypes.string,
+    level: PropTypes.number,
     isBreakpoint: PropTypes.bool,
     toggleBreakpoint: PropTypes.func,
     onClick: PropTypes.func,
@@ -200,6 +201,7 @@ class TestRow extends React.Component {
     //setting component of context menu.
     this.props.setContextMenu(listMenu);
 
+    const index = this.props.index >= 0 ? <span className="index" style={{paddingRight: `${(this.props.level || 0) * 5}px`}}>{this.props.index + 1}.</span> : null;
     const rendered = <tr
       ref={node => {return(this.node = node || this.node);}}
       className={classNames(this.props.className, {"selected": this.props.selected}, {"break-point": this.props.isBreakpoint}, {"dragging": this.props.dragInProgress})}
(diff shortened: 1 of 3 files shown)


Jongkeun:
Do you have any schedule at this pr?
I'm looking forward this commands and `storeEval`.
Let me know the schedule if you have?

Jongkeun:
I've already used flow control on my old customized [selenium.](https://drive.google.com/open?id=1EG0NRXsD5W4kOKBzTK2tqYEbUWCqI_Yx)
I used [sideflow](https://github.com/73rhodes/sideflow).
Is there something I can help you?

Jongkeun:
I already made [flow control](https://drive.google.com/file/d/1pk_RCv5el7O5Gz23fJ7x4dIjx5F1NGm7/view), but not yet done. And I even made `gotoif` and `label`.
Absolutely your result is better than me.
The reason that I made separately is I needed it early. 

Additionally we need 'gotoif' and 'label'.
Because we should make it run old scripts on the new selenium-ide.
Although remove `gotoif`, that is not matter if `gotoif` is migrated to `if` automatically.

But I agree that `eval` has a problem too.
Nevertheless we need the command.
I knew `eval` is not recommend but I am using it by changing extension policy.
Also Katalon is using these command.
Because they have known it is important to user.
I want you to consider it and if you don't want it, think about alternative.

As one of user, I want new selenium-ide to be the best program of several selenium-ide programs.
I would help you as possible as I can.

corevo:
I didn't know you we're working on this, looks amazing.  
About `goto` and `label` we have [this](https://github.com/SeleniumHQ/selenium-ide/pull/139), it will convert `goto` and `label` to `while` and `if`.  
About `eval`, don't get me wrong, I want Selenium IDE to be the best thing out there, but as of this moment, Mozilla sent us a message demanding to remove it.  
After we remove the remnants of the old `eval` system, we'll talk with them about introducing a safer one.  
I am unsure to what agreement Katalon has arrived with Mozilla about unsafe code injections, but it is not something Mozilla have discussed with us.

corevo:
@Jongkeun I was thinking for a bit, why did you change the security policy?  
Do you `eval` the conditions inside the background script?  
You know that would enable access to chrome.debugger and in extension to any origin request?  
There's a huge potential security risk in doing this.

Jongkeun:
@corevo  So I restored policy.
And I changed the way from calling `eval` directly to calling to use `sendMessage` to `commands-api`.
I put `doEval` at `selenium-api.js`.
There are many reasons for the change.
But the important reason is I have to use `storedVars`.
So if I wanna use `eval` I am calling `sendMessage' with eval's condition by using the `ext-command`.
And I get the `returnValue` about the condition.

Anyway, my flow control commands are at extension side. 
Because I should handle `TestCase`.
I couldn't handle `TestCase` on browser side.

What are you worried about?
Do you think the user will become an attacker? Or will  be a victim?

Jongkeun:
When I called eval on browse side, there was no error or warning.

corevo:
I think the user may well become a victim.  
I think it's highly plausible to see someone as a joke tell a user, put this line of code here, and cause a mess, or worse.  
More than that, changing the csp will show unfriendly alerts to the users, which I'd like to avoid.

Jongkeun:
what kind of alerts or warning?
I didn’t have any alert even if I didn’t change csp.

corevo:
Once you publish an extension to the stores, it'll automatically be disabled for all the users.
Saying that the CSP had changed.

Cheers,
Tom

Jongkeun:
If the debug and the publish are different, that is problem as you said. 
I didn’t know about that.

PR #142, opened 2018-04-10
Adding a drag drop feature on TestTable.

Jongkeun:
There was something unfortunate about it.

- I could not get information of file until dropping. So I couldn't use `canDrop` attribute.
filedndwarning
- [react-dropzone](https://github.com/react-dropzone/react-dropzone/) is possible to classify using mime-type. But this can't get and distinguish mime-type of `.side`. So I failed.

I didn't have a choice but to add a warning message after dropping the file.
This is [video](https://drive.google.com/file/d/1Mw711G6WrLxCFwBSnfIiIDZQnD-s-5lm/view).

### DIFF
--- a/packages/selenium-ide/src/neo/components/TestTable/index.jsx
+++ b/packages/selenium-ide/src/neo/components/TestTable/index.jsx
@@ -51,7 +51,7 @@ export default class TestTable extends React.Component {
     removeCommand: PropTypes.func,
     swapCommands: PropTypes.func,
     clearAllCommands: PropTypes.func,
-    accepts: PropTypes.arrayOf(PropTypes.string),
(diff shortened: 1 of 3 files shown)


corevo:
I just pushed [this](https://github.com/SeleniumHQ/selenium-ide/commit/997fad0875e3684b958862a8478600f3c8287483)

corevo:
I know, I'm working on adding the blue shadow.
In any case I had 2 issues.
The drop target was the table rather than the whole app.
Secondly, react-dnd is a low level library, I added react-dropzone which let me do it with a few lines of code, and place it in a self contained component, it doesn't add to the complexity of others that are already overly complex, like the test table.

Cheers,
Tom

PR #143, opened 2018-04-10
if attribute 'comment' does not exist, fill comment value by empty

Jongkeun:
If you used old `.side`, there is no `comment` of `Command`,
So comment was still left even if I load the new project.
This is not necessary, but I think it would be nice.

### DIFF
--- a/packages/selenium-ide/src/neo/models/Command.js
+++ b/packages/selenium-ide/src/neo/models/Command.js
@@ -80,7 +80,7 @@ export default class Command {
 
   static fromJS = function(jsRep) {
     const command = new Command(jsRep.id);
-    command.setComment(jsRep.comment);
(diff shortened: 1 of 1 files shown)


corevo:
I'm aware of that bug, its fixed [here](https://github.com/SeleniumHQ/selenium-ide/pull/139/files#diff-5fa5b430a8a2ad85cba0e61e4d0b1b50R46).  
I found it too, and its also possible to happen in other fields as well if for some obscure reason they are undefined.

sbranston:
The current version of Selenium IDE does not show options for Insert Comment using Firefox Quantum. This was noted as closed, but is still an issue. Can you assist on installing the fix listed above?

Jongkeun:
Hi @sbranston,
That is not merged yet.

PR #144, opened 2018-04-10
Recorder - migrating DOMNodeInserted to MutationObserver

grese:
# Summary:
The onDOMNodeInserted event is deprecated, and throws a warning.
Replacing usage of onDOMNodeInserted event in the 'Record' module
with a MutationObserver.

#### Notes:
There are more instances of mutation events in selenium-api.  If possible,
it would be nice to move those to mutation observers as well.

### DIFF
--- a/packages/selenium-ide/src/content/record-api.js
+++ b/packages/selenium-ide/src/content/record-api.js
@@ -27,6 +27,7 @@ function Recorder(window) {
 }
 
 Recorder.eventHandlers = {};
+Recorder.mutationObservers = {};
 Recorder.addEventHandler = function(handlerName, eventName, handler, options) {
   handler.handlerName = handlerName;
   if (!options) options = false;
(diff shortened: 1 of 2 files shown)


corevo:
Just putting this [article](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events) here.  
I'll take a look next week.

grese:
@corevo - Cool... thanks.  This was my first stab at contributing to this repo, so let me know if you want it implemented differently.

Do you happen to know why it is necessary to record a "mouseOver" event when "DOMNodeInserted" fires?  The code that I implemented _should_ behave the same as "DOMNodeInserted", but I'd like to make sure that it handles the specific use-case(s) correctly.

corevo:
I'm traveling atm, I'll definitely have a look next work week.

corevo:
Phew, sorry it took me so long to get this in.  
Thanks for contributing 👏

PR #146, opened 2018-04-12
When toggleConsole to show Log, restore size to stored size, not maximize

Jongkeun:
if someone calls `toggleConsole`,  the log form is maximized regardless of user setting
So I modified the part for regard of user setting.

if `consoleSize` of `storage` was same with `minconsoleHeight` when you called `toggleConsole`, 
I made one log line to be shown.

### DIFF
--- a/packages/selenium-ide/src/neo/stores/view/UiState.js
+++ b/packages/selenium-ide/src/neo/stores/view/UiState.js
@@ -188,13 +188,13 @@ class UiState {
     this.resizeConsole(this.minConsoleHeight);
   }
 
-  @action.bound storedsizeConsole() {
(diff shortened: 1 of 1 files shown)


packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -186,9 +188,13 @@ class UiState {
     this.resizeConsole(this.minConsoleHeight);
   }
 
+  @action.bound storedsizeConsole() {
```
corevo: Can you rename this func to `restoreConsoleSize`?

Jongkeun:
I'll go change it at work tomorrow.

PR #152, opened 2018-04-16
fixed #147.

### LINKED ISSUE #147: When I minimize the log form, I see something.
Left is something that I modified, and right is original the last version.

It seems like to happen when there is some log.

I modified `minConsoleHeight` from 30 to 29.
But I agree to 30 is better.

What is the best way  to fix it simply?

Jongkeun:
I fixed #147 issue.

### DIFF
--- a/packages/selenium-ide/src/neo/components/TabBar/style.css
+++ b/packages/selenium-ide/src/neo/components/TabBar/style.css
@@ -1,6 +1,6 @@
 .tabbar {
   background-color: #F3F3F3;
-  padding: 0 5px 1px;
+  padding: 0 5px;
   position: relative;
 }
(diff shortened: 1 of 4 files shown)


corevo:
I don't like this fix because we assume the console will be rendered at exactly 29px, while we hardcoded the console height [here](https://github.com/SeleniumHQ/selenium-ide/blob/master/packages/selenium-ide/src/neo/stores/view/UiState.js#L34).  
What I think we should do, is make sure `.tagbar` is exactly 30px high, not by padding, but by literally set its height to that, and then render all the children in relation to that.

Jongkeun:
What do you think of this way?

PR #167, opened 2018-04-24
There is error message when log tabbar click.

Jongkeun:
fixed error message when I click the `TabBar`.
And I thought that open the `TabBar` by clicking when `TabBar` was minimized is better.

### DIFF
--- a/packages/selenium-ide/src/neo/containers/Panel/index.jsx
+++ b/packages/selenium-ide/src/neo/containers/Panel/index.jsx
@@ -176,7 +176,9 @@ modify(project);
                 </SplitPane>
               </div>
             </div>
-            <Console height={UiState.consoleHeight} />
+            <Console
+              height={UiState.consoleHeight}
+              restoreSize={UiState.restoreConsoleSize} />
           </SplitPane>
           <Modal project={this.state.project} />
           <Tooltip />
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "pack:firefox": "cd packages/selenium-ide && yarn build-firefox",
     "test": "jest",
     "test:webdriver": "cd tests/webdriver && node ../../packages/selenium-side-runner/dist/index.js",
-    "lint": "eslint packages/selenium-ide/src/{neo,selianize}/ packages/selenium-side-runner/src/ --ext .js --ext .jsx",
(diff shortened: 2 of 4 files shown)


corevo:
Nice fix, I also like the idea of restoring the Console if clicked on it.  
But, for higher reusability, I don't want the restore height to be in the `TabBar` component, not all tab bars are draggable.  
Instead of having this prop, restore the height on the tabChanged callback in the Console, better yet add an onClick handler that will fire on every tab click and restore it on that.

Jongkeun:
Oh I understand what you are saying.
`TabBar` is used by other components too.
I will try to fix it.

Jongkeun:
My think was wrong... Anyway, I'll fix it.

Jongkeun:
I don't think I exactly understood it.
I thought you told me to use `tabChanged` callback. 
But tabChanged event was not broken.
because of [this](https://github.com/SeleniumHQ/selenium-ide/pull/167/files#diff-c27b812835828c93e0a4fdca8e17d135L48) condition.

So I judged that it is impossible to use.
And I modified something.
I think it is not bad.😅
Could you tell me easily?

packages/selenium-ide/src/neo/containers/Panel/index.jsx
```diff
@@ -176,7 +176,9 @@ modify(project);
                 </SplitPane>
               </div>
             </div>
-            <Console height={UiState.consoleHeight} />
+            <Console
+              height={UiState.consoleHeight}
+              restoreSize={UiState.restoreConsoleSize} />
```
corevo: The prop is not specified as required, but you're not checking if it's provided or not.  
Either specify it as required, or check that it's not null.

corevo:
It's a good solution, but what I had in mind is instead of putting the onClick on the entire footer, add an onClick callback to the TabBar and then the Console will listen to that, and restore the console size

Jongkeun:
Is this your intention?
I modified like the `Navigation`.
But it is not working. 
Because it is not `tabChanged` as I told you.

Jongkeun:
I have thought `tabChanged` and `tabClicked` must be different.
So, although you kept saying, I didn't use `tabChanged`.

And I  don't think it's necessary to check `restoreSize` prop.
This is just function. Why should I check if it is provided? 
Let me know please.
Thanks.

PR #169, opened 2018-04-25
fixed error message when someone drags testrow to outside

Jongkeun:
How to reproduce.
 - Just drag second testrow to last.

image

This is a bug of `dropzone`.
I didn't create issue about this problem.
This is workaround.
I got this message from selenium-ide.

image

So I searched for the cause in `dropzone`.
I can find this [code](https://github.com/react-dropzone/react-dropzone/search?utf8=%E2%9C%93&q=createObjectURL&type=).

### DIFF
--- a/packages/selenium-ide/src/neo/components/SuiteDropzone/index.jsx
+++ b/packages/selenium-ide/src/neo/components/SuiteDropzone/index.jsx
@@ -12,7 +12,7 @@ export default class SuiteDropzone extends React.Component {
   }
   render() {
     return (
-      <Dropzone className="suite-dropzone" acceptClassName="accept" accept=".side" disableClick={true} multiple={false} onDropAccepted={this.onDrop.bind(this)}>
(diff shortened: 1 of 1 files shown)


Jongkeun:
What's wrong?
Do you need me to do something?
image

corevo:
No, node 10 was released yesterday, and our dependencies are not compatible with it.  
I'll fix it.

Jongkeun:
`disablePreview` has no effect on this problem.
The problem of smooth was started from adding `dropzone`.
We didn't use `disablePreview` attribute. didn't you?
If you test using released selenium-ide, you can find same problem.
Do I misunderstand?

corevo:
Fixed both issues with [this](https://github.com/SeleniumHQ/selenium-ide/commit/469be3218a6a0d6cee6881a3f5b19f7571116034).  
I really didn't realize the two would interfere, since it's not a native file.

Issue #160, opened 2018-04-17
Request for command storeAttribute 


jngale:
Would it be possible to add these commands?  Thank you

I would like to add that there seems to be a large number of commands that are not included yet. I do understand priorities, but it is getting challenging. I have just run into a need for storeAttribute and storeXpathCount. Perhaps it is just me always needing what isn't there yet. Thanks so much for all your work and help

I am presently using the latest Chrome version.

jngale:
I would like to add that there seems to be a large number of commands that are not included yet. I do understand priorities, but it is getting challenging. I have just run into a need for storeAttribute and storeXpathCount.  Perhaps it is just me always needing what isn't there yet.  Thanks so much for all your work and help

corevo:
`assertSelectedLabel` and `verifySelectedLabel` were added in [3.1.0](https://github.com/SeleniumHQ/selenium-ide/releases/tag/v3.1.0)

corevo:
Fixed in #171

PR #172, opened 2018-04-30
Add assertSelectedLabel and verifySelectedLabel

Plotist:
By request from https://github.com/SeleniumHQ/selenium-ide/issues/160


Plotist:
@corevo is there something that needs to be done for it to be merged ?

corevo:
Since behind the scenes when we export `verify` and `assert` are the same, the test is needlessly slower.  
Instead change the selected label, and check that it works.

PR #204, opened 2018-05-23
Show deprecated message at Wait... commands 

Jongkeun:
As the user's point of view, I felt confused by the sudden disappearance of the `wait..` commands.
Of course it doesn't matter if he knows why doesn't need those anymore, but I don't think that they knew.
So I thought it was better to tell and show them commands were deprecated.

### DIFF
--- a/packages/selenium-ide/src/content/commands-api.js
+++ b/packages/selenium-ide/src/content/commands-api.js
@@ -68,6 +68,9 @@ function doCommands(request, sender, sendResponse) {
           document.body.removeAttribute("SideeXPlayingFlag");
           sendResponse({result: e.message});
         }
+      } else if (request.commands.toLowerCase().includes("wait")) {
+        selenium["doDeprecated"](request.target, selenium.preprocessParameter(request.value));
+        sendResponse({result: "deprecated"});
       } else {
         sendResponse({ result: "Unknown command: " + request.commands });
       }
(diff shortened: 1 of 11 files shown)


Jongkeun:
Have you any schedule about this error?
(`componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops  react/no-deprecated`)
I saw it when I called the `yarn lint` command.

corevo:
First off, wait commands aren't deprecated, they are obsolete, the difference being that deprecated code still works but puts out errors, while obsolete code errors out and puts errors.  
The correct behavior is not adding a deprecation warning, but erring the test case (the status should be error, and the description should say that they are obsolete).  
Now to the more difficult problem, wait commands or any command that is not recognized, will stop the IDE from saving the project, because it will fail when executing the test case from command line.  
Removing the waits automatically will lift the burden from the users of doing it manually (they may have dozens).  
Perhaps the best solution is still removing the waits automatically, putting out a dialog if any wait was removed saying that waits are obsolete, and showing errors in case a wait was used.

Jongkeun:
No I just upgrade my node version.😄
Never mind. It doesn't matter to me.

I agreed with you.
However, we should tell users that wait commands are no longer necessary via any ways.
If you don't, there will be many issues created by user.
Because they didn't know.

corevo:
We should notify them, except the error should be different, instead of unknown command we should say it's obsolete

Cheers,
Tom

PR #220, opened 2018-05-31
Add StoredVars tab

Jongkeun:
I need the `storedVar` tab before reference tab.


Jongkeun:
I knew there is many things to fix at your sight.😂
Please let me know about those.

corevo:
That's an awesome addition 😲  
I'm sure theres things that needs fixing 😱  
Jokes aside, let's get this merged!

packages/selenium-ide/src/neo/components/TabBar/index.jsx
```diff
@@ -52,7 +52,7 @@ export default class TabBar extends React.Component {
       });
       if (this.props.tabChanged) this.props.tabChanged(tab);
     }
-    if (this.props.tabClicked) this.props.tabClicked(tab);
+    if (this.props.tabClicked) this.props.tabClicked();
```
corevo: Even though for most cases `tabChanged` is the better event, `tabClicked` may also be useful.  
And it's user should know which one was clicked, especially if he doesn't listen to `tabChanged`.

packages/selenium-ide/src/neo/components/StoredVarList/style.css
```diff
@@ -0,0 +1,40 @@
+.valHeader {
```
corevo: css is case insensitive, when you define a class name keep it verbose and hyphen cased: `value-header`.  
Maybe I should add style linting.

packages/selenium-ide/src/neo/components/StoredVarList/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+import React from "react";
+import PropTypes from "prop-types";
+import StoredVar, { StoredVarAddBtn } from "../StoredVar";
+import { getStoredVars, setStoredVar, deleteStoredVar } from "../../IO/SideeX/formatCommand";
+import "./style.css";
+
+export default class StoredVarList extends React.Component {
+  constructor(props){
+    super(props);
+    this.state = { add: false };
```
corevo: The state is terse and not descriptive of it's use, perhaps calling it `this.state.addingVariable` will make it clearer.

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+import "./style.css";
+
+export default class StoredVar extends React.Component {
+  constructor(props){
+    super(props);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+    this.keyChanged = this.keyChanged.bind(this);
+    this.valueChanged = this.valueChanged.bind(this);
+    this.saveValue = { key: "", value: "" };
```
corevo: This is a `state`

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+  handleChange(e) {
+    this.props.edit(this.props.keyVar, e.target.value);
+  }
+  delete(){
+    this.props.delete(this.props.keyVar);
+  }
+  keyChanged(e) {
+    this.saveValue.key = e.target.value;
```
corevo: `setState` here

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+    this.saveValue.value = e.target.value;
+  }
+  render() {
+    return (
+      <div className="row">
+        <div className="cell index">
+          {this.props.index} {this.props.add ? "" : "."}
+        </div>
+        <div className="cell storedKey">
+          {this.props.add ? <input type="text" onKeyDown={this.handleKeyDown} onChange={this.keyChanged}/> : this.props.keyVar}
```
corevo: Pass it down here as value prop

packages/selenium-ide/src/neo/IO/SideeX/formatCommand.js
```diff
@@ -70,5 +70,23 @@ function handleFormatCommand(message) {
   }
 }
```
corevo: The biggest problem here, is that `declaredVars` is neither a `state` nor an `observable`.  
Which means that when changed React can't tell wether to render or not, more specifically, react doesn't care if it gets changed.  
So instead theres `this.state.refresh`, but that completely overrides reacts lifecycle and optimizations, by telling react, "trust me, you need to render".  
Instead create a `Variables` store, and make the variables an [Observable Map](https://mobx.js.org/refguide/map.html).  
Have a look at my upcoming refactor to the [Logs](https://github.com/SeleniumHQ/selenium-ide/blob/extensibility/packages/selenium-ide/src/neo/stores/view/Logs.js)

corevo (changes requested):
I really, really like the idea, but it's inefficient at rendering, I wrote you some key pointers.

packages/selenium-ide/src/neo/components/TabBar/index.jsx
Jongkeun: I removed this because I received the tab parameter, but I didn't use at this case.
So I worried the warning from the `yarn lint`.
I agree with your idea. I am going to add soon.

packages/selenium-ide/src/neo/components/StoredVarList/style.css
Jongkeun: `css` is my the worst point about web programming.
I worked at web programmer, but I've never touch the UI of the web page.
I might fix again and again.

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
Jongkeun: I didn't wanna make this `state`.
There is some reason.

1. stored the value '123' of the key 'abc'
2. stored the value '456' of the key 'abc2'

In this case, while typing the 'abc2', you are typed 'abc'. 
So if I use `state`, I end up saving unwanted values at the key 'abc'.

What do you think about it? @corevo .

packages/selenium-ide/src/neo/IO/SideeX/formatCommand.js
Jongkeun: I considered about the `Observable` and the ` Store`.
But I decided like this concept.
Because while running a `testcase`, if you wanna edit the value, but you can't by changed the value fast. 
The plugin which I used didn't supported to auto refresh either. 
I thought it has the reason about not supporting auto refresh.

I am concerned about the problems that occur when I make this Observable.

packages/selenium-ide/src/neo/components/StoredVarList/style.css
corevo: Don't worry about it 😉

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
corevo: I'm sorry, I did not understand your point

packages/selenium-ide/src/neo/IO/SideeX/formatCommand.js
corevo: I think I have a good idea around it,  
Just like when debugging in Chrome, you can only edit the values when hitting a break point.  
Don't enable editing unless on a break point.

Jongkeun: Such a good idea.
I will try it.

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
Jongkeun: Sorry. 
Whenever I call `keyChanged`, I thought it was saved.
I fixed it.

packages/selenium-ide/src/neo/IO/SideeX/formatCommand.js
Jongkeun: @corevo, If I am going to try to use Observable, it is possible when `formatCommand` is class as far as I know. 
May I change the `formatCommand` to class?
Or is there another way which I don't know?

corevo: No, create a new view store, called Variables.

Jongkeun:
It's finished for me.
I didn't add what we mentioned about a break point.
Because I thought it makes the code complex and it's fine when I used.

There is a problem.
Clear button's tooltip is not working properly.
If I click the other tab of `console`, tooltip is disappeared.
And appears after you run a script.
Is it `react-tooltip`'s bug? or I can fix by changing my code?

Jongkeun:
@corevo While merging the last code, I saw you declared `Logger` at `formatCommand`.
I am going to move my code to `formatCommand` too.
Then I don't have to add `addListener` again.

packages/selenium-ide/src/neo/containers/Console/index.jsx
```diff
@@ -19,21 +19,22 @@ import React from "react";
...
 import CommandReference from "../../components/CommandReference";
+import variables from "../../stores/view/Variables";
 
 export default class Console extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
-      tab: "Log"
-    };
+    this.state = { tab: "Log", refresh: 0 };
```
corevo: Is refresh necessary?

Jongkeun: Sorry. I am going to remove after having a dinner.

corevo: Great, looks amazing so far, just get rid of the button and the glyph too.

packages/selenium-ide/src/neo/stores/view/Variables.js
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { action, observable } from "mobx";
+
+class Variables {
```
corevo: This looks really good, let's test this.  
Add a spec file here `packages/selenium-ide/src/neo/__test__/stores/domain/view/Variables.spec.js`.  
Have a look [here](https://github.com/SeleniumHQ/selenium-ide/blob/master/packages/selenium-ide/src/neo/__test__/stores/domain/ProjectStore.spec.js) for ideas.

Jongkeun: I did.
Let me know if it has correction points.

packages/selenium-ide/src/neo/__test__/stores/domain/view/Variables.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+  it("should add new key and value to the map", () => {
+    Variables.clearVariables();
+    const key = "key1";
+    const value = "value1";
+    Variables.addVariable(key, value);
+    expect(Variables.storedVars.size).toBe(1);
+  });
+  it("should not add duplicates to the map", () => {
+    Variables.clearVariables();
```
corevo: Use [afterEach](https://facebook.github.io/jest/docs/en/api.html#aftereachfn-timeout) for this

packages/selenium-ide/src/neo/__test__/stores/domain/view/Variables.spec.js
```diff
@@ -0,0 +1,63 @@
...
+// under the License.
+
+import { useStrict } from "mobx";
+import Variables from "../../../../stores/view/Variables";
+
+useStrict(true);
+
+describe("Varaibles", () => {
+  it("should clear the map", () => {
+    Variables.clearVariables();
```
corevo: This test checks that that `clearVariables` is noop, add a variable beforehand.

packages/selenium-ide/src/neo/__test__/stores/domain/view/Variables.spec.js
```diff
@@ -0,0 +1,63 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { useStrict } from "mobx";
+import Variables from "../../../../stores/view/Variables";
+
+useStrict(true);
+
+describe("Varaibles", () => {
```
corevo: "Variables"

packages/selenium-ide/src/neo/components/StoredVarList/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+// under the License.
+
+import React from "react";
+import PropTypes from "prop-types";
+import StoredVar, { StoredVarAddBtn } from "../StoredVar";
+import { observer } from "mobx-react";
+import "./style.css";
+
+@observer
+export default class StoredVarList extends React.Component {
```
corevo: I'd like to change the component's names
StoredVarList -> VariableList
StoredVar -> Variable
For the same reasons we changed the tab name, the implementation should be consistent in naming, for the sake of readability

packages/selenium-ide/src/neo/components/StoredVar/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+        </div>
+        <div className="cell col">:</div>
+        <div className="cell storedVar">
+          {this.props.add ? <input type="text" onKeyDown={this.handleKeyDown} onChange={this.valueChanged} /> :
+            <ContentEditable className="value" onChange={this.handleChange} html={this.props.value} onKeyDown={this.handleKeyDown} />}
```
corevo: I don't like the fact that it's not a form, because it makes you check that the user pressed `Enter`, rather than use `onSubmit`.  
I'll have to check the behavior manually to make sure it feels fine without a form, I'll let you know.

Jongkeun: Thanks.

corevo:
Some visual changes:  
- Remove the row numbers, its unhelpful.
- Move the plus next to the clear button.
- Don't zebra the table.
- Allow changing the variable name.
- Don't let users edit anything unless they are on a breakpoint (it may cause unintentional side effects).
- Align everything to the left, nothing in the IDE is centered aligned it feels alien.
- Clicking outside is confirming, in general I want the same feeling as editing CSS in Chrome's devtools.

corevo:
I want `packages/selenium-ide/src/neo/stores/view/Variables.js` for #250, do you think you could extract it from this PR and post it, so I can merge it?

Jongkeun:
Is there any way I can do it other than new PR?
I thought a way is that I create new PR.

corevo:
Send a new one, if it's too much trouble then dont worry about it

Cheers,
Tom

Jongkeun:
I will do tomorrow.


네이버 메일앱에서 보냈습니다.
-----Original Message-----
보낸사람: "Tomer Steinfeld"<notifications@github.com>
받는사람: "SeleniumHQ/selenium-ide"<selenium-ide@noreply.github.com>
참조: "Jongkuen Hong"<dldkfod@naver.com>, "Author"<author@noreply.github.com>
날짜: 2018. 6. 18 오전 12:08:50
제목: Re: [SeleniumHQ/selenium-ide] Add StoredVars tab (#220)

Send a new one, if it's too much trouble then dont worry about it

Cheers,
Tom

packages/selenium-ide/src/neo/stores/view/Variables.js
```diff
@@ -0,0 +1,53 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { action, observable } from "mobx";
+import PlaybackState from "./PlaybackState";
+
+class Variables {
+  @observable storedVars = new Map();
+  @observable addingVariable = false;
```
corevo: This should be internal state of the Variables component, I removed it from #253

packages/selenium-ide/src/neo/stores/view/Variables.js
```diff
@@ -0,0 +1,53 @@
...
+  @observable isEditable =  (!PlaybackState.isPlaying || PlaybackState.paused || PlaybackState.isStopping);
```
corevo: Make this a [@computed](https://mobx.js.org/refguide/computed-decorator.html) property, a value that is inferred from other observables should be a computed, it's more efficient.  
Also it should be `(!isPlaying || paused)`, `isStopping` should not let you edit.

Jongkeun:
I moved the plus button next to the clear button on my local PC.
But I should add the state and function for `Add Variable`.
Because `VariableList` have to know if the plus button clicked or not.
I hope that I don't wanna put the state and the function on `Console`.
I think it is better and clear that the function about the plus button is located in `VariableList` 
What do you think about?

corevo:
I understand what you're saying, I don't wanna wire it up in a weird way.  
`Console` is a container, I'm fine with it having a `this.state.isAddingVariable`.

Jongkeun:
I'm done.
Check it please.

Jongkeun:
Please get me the design shot which you want when you have a time.

corevo:
I don't have a screenshot of how this supposed to look, I want it to speak the same design language as the rest of the IDE.  
For example, nowhere we separate values with `:`, and the inputs are very different than the rest of the IDE (e.g. no padding, different borders).

Jongkeun:
1. remove the `ContentEditable" -> **OK**
2. change `div` to `ul` and `li`. -> **OK**
3. change the css name `.row` to `.variable` etc. -> **OK**
4. remove `:` -> **OK**
5. make it felt like it's an organic part of the IDE -> **Tried**.
6. make this to `form`. -> **OK**

I made this to form, but I don't know about the reason.
I don't like the wide view of `Variables`, but I had no choice to make for 5.
Please give me some advice.

Jongkeun:
@corevo When will this be merged?

corevo:
I wasn't aware if it's still under development or not, I'll make sure to have a look.

corevo:
This looks very promising on the GUI side of it, just a few things.
 - If I'm editing keep the input background white on hover the rest can stay gray.
 - Keep an invisible border around the inputs at all time, because you change the input radius the text jumps a bit and it irks me.
- Keep the padding consistent with the rest of the IDE (I think it is `padding: 4px 5px;` check the `CommandForm` to be sure)
- Add margin around the inputs, It doesn't look good that the input's border borders with the background (1 or 2 pixels should suffice).  

I'll have a look at the code later on, great work! 💯

Jongkeun:
I modified that you mentioned.
Check it, please.

packages/selenium-ide/src/neo/components/Variable/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+      e.preventDefault();
+      const isValidKey = this.state.key || this.props.keyVar;
+      const isValidValue = this.state.value || this.props.value;
+      if (isValidKey && isValidValue) {
+        this.edit(isValidKey, isValidValue);
+      }
+    }
+  }
+  handleChanged() {
+    const isValidKey = this.state.key || this.props.keyVar;
```
corevo: The naming convention is misleading, variables that begin with `is` are booleans, while these are strings.  
Perhaps call them `validKey`?

packages/selenium-ide/src/neo/components/Variable/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+    const isValidValue = this.state.value || this.props.value;
```
corevo: Why is empty string considered invalid? users can definitely store an empty string.

packages/selenium-ide/src/neo/containers/Console/index.jsx
```diff
@@ -58,24 +60,40 @@ export default class Console extends React.Component {
...
+  //create different object which stores name and read status (e.g., unread boolean)
+  tabClicked(tab) {
+    this.setState({
+      tab
+    });
+    if(this.props.restoreSize) this.props.restoreSize();
+  }
+  addVariableClicked(isAdding) {
+    this.setState({
+      isAddingVariable: isAdding
```
corevo: I don't like this state, it is also inconsistent with the experience we give with the command table,  
let's just have a "pristine" variable line, just like in the command table.

corevo (changes requested):
Some general feedback:
 - I have to click twice to change between the variable name to it's value, once to blur the previous input, and another to focus the next, instead of simply moving between them.
 - Hitting tab won't switch between the variable name and value inputs.
 - Hitting tab also locks you inside the variable editing and won't let you reach other elements.

Jongkeun:
@corevo Could you check it out?

corevo:
I was sick the past few weeks, I'll make sure to have a look.

Jongkeun:
I don't care what happens to this pr, even though this pr was deleted.
This pr is in your hands. I will just follow your think.
Let me know if there's anything I need to do.

cdifino:
Hi @corevo ! I would like to know what happened to this PR? What is the status? and is someone working on it? I think it is a great feature to be implemented in the extension.

corevo:
@cdifino the problem is typings, since `execute script` may return for example an array, this variables tab will cast it to string.

We need to support specific forms for specific variable types as to avoid casting them to strings.

PR #223, opened 2018-06-04
console height bug fix when changing window size with console minimized

Jongkeun:
Test case.

1. make the selenium window maximize and the console maximize.
2. folding console by clicking ESC.
3. make the selenium window to original size.
4. And click ESC for restoring console size.

when you have this scenario, had d bug about console.

### DIFF
--- a/packages/selenium-ide/src/neo/stores/view/UiState.js
+++ b/packages/selenium-ide/src/neo/stores/view/UiState.js
@@ -173,8 +173,12 @@ class UiState {
   }
 
   @action.bound resizeConsole(height) {
+    var maxConsoleHeight = this.windowHeight - this.minContentHeight;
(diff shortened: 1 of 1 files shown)


packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -173,8 +173,12 @@ class UiState {
   }
 
   @action.bound resizeConsole(height) {
+    var maxConsoleHeight = this.windowHeight - this.minContentHeight;
```
corevo: I'm unsure why travis didn't fail the build for it, since `no-var` rule is set.  
Change this to `const`.

Jongkeun: Sure I did.

PR #227, opened 2018-06-08
fix: regex to parse json capability options in side-runner cli

toshiya:
Hi, I'm a test automation engineer in Japan. The Selenium IDE project greatly helps our test automation process. I appreciate all the project members.

I found a small fix in selenium-side-runner (Maybe not intended behavior) in current master branch.
Please check it if you have time. 

Thanks.

# What is this ?

- fix parse logic in selenium-side-runner to consider nested json strings.

Our use case is to use `chromeOptinos.mobileEmulation` in selenium-side-runner.
(IDE Recording is fine for mobile emulated browser.)

Example commands are as below. Also, I added a test case.

```
cd ./packages/selenium-side-runner
yarn build
./dist/index.js -c 'browserName=chrome chromeOptions.mobileEmulation="{"deviceName" : "iPhone 6"}"'  some-project.side
```

Currently, the above strings cause JSON parse error in packages/selenium-side-runner/src/capabilities.js because of `"`s are not trimmed properly.

This PR fix it.

# Our Environment

- Nodejs: v8.11.1
- macOS 10.13.5 (High Sierra)
- original selenium-ide commit hash: a51f9679872905a035ea22d7b5d968c8c652f0d2

### DIFF
--- a/packages/selenium-side-runner/src/__test__/capabilities.spec.js
+++ b/packages/selenium-side-runner/src/__test__/capabilities.spec.js
@@ -44,6 +44,16 @@ describe("capabilities string parser", () => {
       binary: "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
     });
   });
+  it("should parse capability key with nested json", () => {
+    const capabilities = "chromeOptions.mobileEmulation=\"{\"deviceName\" : \"iPhone 6\"}\"";
+    expect(Capabilities.parseString(capabilities)).toEqual({
+      chromeOptions: {
+        mobileEmulation: {
+          deviceName: "iPhone 6"
+        }
+      }
+    });
+  });
   it("should parse boolean capability key", () => {
     const capabilities = "javascriptEnabled=false databaseEnabled=true";
     expect(Capabilities.parseString(capabilities)).toEqual({
--- a/packages/selenium-side-runner/src/__test__/capabilities.spec.js
+++ b/packages/selenium-side-runner/src/__test__/capabilities.spec.js
@@ -44,16 +44,6 @@ describe("capabilities string parser", () => {
       binary: "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
     });
   });
-  it("should parse capability key with nested json", () => {
-    const capabilities = "chromeOptions.mobileEmulation=\"{\"deviceName\" : \"iPhone 6\"}\"";
-    expect(Capabilities.parseString(capabilities)).toEqual({
-      chromeOptions: {
-        mobileEmulation: {
-          deviceName: "iPhone 6"
-        }
-      }
-    });
-  });
   it("should parse boolean capability key", () => {
     const capabilities = "javascriptEnabled=false databaseEnabled=true";
     expect(Capabilities.parseString(capabilities)).toEqual({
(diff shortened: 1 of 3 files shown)


corevo:
I didn't expect people to pass inline stringified JSONs.  
In your case I envisioned  
```bash
selenium-side-runner -c 'browserName=chrome chromeOptions.mobileEmulation.deviceName="iPhone 6"' demo.side
```  
It felt to me that passing JSONs inline like that is very uncomfortable, so I've come up with this serialization mechanism.  
I'll have to think about this, because even though this isn't my intended behavior, perhaps not having this will cause even more confusion.

toshiya:
Thank you for your explanation. I understood your intension.

Although current key-value based notation is good for simple capabilities, I think it's more convenient to use JSON strings directly for large and complex capabilities. 

My proposal is adding another cli option to specify [configuration](https://github.com/toshiya/selenium-ide/blob/57bbf3e3fea6b6397fa4fb07c28e8cb603a43af8/packages/selenium-side-runner/src/index.js#L56-L63) directly by a json file. The configuration variable would be overwritten with a given JSON.

For my above example, it would be

```
$ cat my-configuration.json
{
  "capabilities": {
    "browserName": "chrome",
    "chromeOptions": {
      "mobileEmulation": {
        "deviceName": "iPhone 6"
      }
    }
  }
}
$ ./dist/index.js --configuration-file my-configuration.json demo.side
```

I can implement the option if you like it.

toshiya:
I noticed that .side.yml is exactly the same option what I described above.

corevo:
Exactly, you can add it via a YAML file.  
Do you think JSON is necessary?

toshiya:
Thanks for the reply.

I think supporting YAML only is enough. It's not necessary to support JSON now.
Instead,  it's better to have getting started page for the side-runner.

I could make getting started in another PR. But before that, I  would like to know your plan for user documentation.

corevo:
I've got [this](https://github.com/SeleniumHQ/selenium-ide/tree/master/packages/selenium-side-runner) going on, but I'm not sure how descriptive I need to be, or what information will most likely be what users would go for.  
If you wanna have a look and give me feedback over it, I'll greatly appreciate it.

corevo:
Maybe adding the `--configuration-file` flag is a good idea, it can be used to override the `.side.yml` path.

toshiya:
Thank you for the link.

## Feedback to README

When I first read the document, I didn't fully understand the meaning and acceptable format of each option.

After reading the code of selenium-side-runner, I found out that each option will be passed to the options of [this npm module](https://github.com/applitools/jest-environment-selenium). Then, I understand what each option meant.

I think it's useful to add a brief explanation of internal implementation. Generating temporal test files and run these tests with jest... such things. It greatly helps users to dig into the tool by their selves.

## About `--configuration-file` flags

> Maybe adding the --configuration-file flag is a good idea, it can be used to override the .side.yml path.

I added this flag in this PR. Please check it when you have time.
The flag behavior will be

- with `--configuration-file`, use the specified file to overwrite configuration
- without `--configuration-file`, use  `.side.yml` if it presents and use nothing `.side.yml` doesn't presents. (I removed --no-sideyml flag because this flag seems to be redundant.)


Anyway, I'm looking forward the future of this Selenium IDE projects.

Thanks,

corevo:
Looks great, I just would rather not go too deep into implementation details.  
I'll try to find the right balance between details and verbosity.

PR #228, opened 2018-06-10
WIP: Allow plugins to define new locator types

corevo:
- [x] Define new type
- [x] Resolve that type
- [x] Export code to support that locator
- [x] Disallow redefinition of default locators (id, name etc...)
- [ ] Allow the plugin to work for locators set in value as well


gwtuser:
Does the latest release (v3.2.x) resolve the remaining problems preventing custom locators?  The WIKI I was directed to: https://github.com/SeleniumHQ/selenium-ide/wiki/Getting-Started-with-Plugins, dated June 19, claims that only a few issues remain.

corevo:
There is an issue with knowing wether a specific command requires a locator in advanced.  
The way the API works behind the scenes, is when a command that requires a locator, has a 3rd party one, instead we redirect the locator resolution to the plugin, which returns a valid `xpath` to it.  
The way we currently know if a command requires a locator is only when the command is actually running, we need to add more extensive type annotating to commands, which we did as part of the reference tab, but not sufficient enough (not only know if we need a locator, but wether a locator is actually valid, or 3rd party).  
Also another problem is recording that locator, the IDE has to propagate all recording events to plugins, so that they can add their own locators to the mix, how do we order them?  
Currently order is done by a common understanding (that isn't always true) that: `id > name > linkText > css > xpath`.

mchirlin:
Is there any work still being done on adding plugins to IDE? I have a use case for adding locators and exporting from the IDE for recording particular web applications, but it depends on being able to define a new locator so I can return particular targets while recording.

Issue #189, opened 2018-05-04
copy/cut/paste multiple commands


jngale:
When I write tests, I often find the need to break apart my complex creations into more manageable chunks. Is there a way to copy/cut/paste multiple commands between tests?  This would be an awesome add.

corevo:
Sounds interesting, but we don't have a way to handle more than one item selected.  
I'll have to think about it, but it definitely sounds like something we should do.

saradhit:
copy/cut/paste multiple commands helps a lot...thanks for looking in to it @corevo

luquitia:
This would be extremely helpful!   Thanks!

On a related note, the new .side file type seems to group all "tests" into one file (unlike the old ide file structure which stored each test as separate .html file), so I can't copy or move individual tests from one "suite" (or "project" .side file) into another, except by manually trying to copy the text from notepad, which is difficult to do with human imprecision).

Being able to select 5+ commands at once to cut or copy and move into another test in same project or even across projects would be very helpful (the latter piece would ameliorate the problem with being unable to copy/move individual tests across suites/projects).

luquitia:
Any update on this?  Even if I could copy multiple lines within same test case, that would be helpful.  Record doesn't capture everything I need and it is labor-intensive to copy/paste one line at a time when I need to reuse 5 commands (or sometimes more) later in the same test case.

cawoodm:
Could really use this feature. I feel Copy and Delete should work on multiple commands for a really effective tool.

mdken:
I also feel that this feature would be a big help when writing tests.

p2635:
Yes please, I really need this!

98gmarquee:
Just checking to see if any headway has been made with this. The more I use the new IDE the more I am really missing the ability to copy/paste steps.

Thanks!

trish1400:
Just another vote for this.  I've just started using the new IDE and I'm really missing this.

DorotaGK:
+1

- copying of multiple commands between test scenarios
- copying full test cases between projects

myusernamedm:
Another vote for this. Could save me a huge amount of time. Thanks.

98gmarquee:
Yeah, this chewed up a lot of time for me again today. 

[...]

PR #241, opened 2018-06-15
Multi select (#189)

Jongkeun:
#189 
- TestRow
  - [x] multi-select by ctrl
  - [x] multi-select by shift
  - [x] select all short-cut
  - [x] multi-copy
  - [x] multi-paste
  - [x] sort by index, not click sequence


packages/selenium-ide/src/neo/components/TestRow/index.jsx
```diff
@@ -183,8 +183,18 @@ class TestRow extends React.Component {
   paste() {
     this.props.pasteFromClipboard(this.props.index);
   }
-  select() {
-    this.props.select(this.props.command);
+  select(e) {
+    if (e.ctrlKey) {
```
corevo: Thats not gonna work on macOS because the correct modifier key for mac is `metaKey`.  
Look at the code in `handleKeyDown` and use `primaryKey` instead.

Jongkeun: Now I classify `onFocus` and `onClick`. 
But still I can't find the way to work appropriately. 😨 
I realized why `onFocus` is need.
I'll make it fixed by referenced your comment.
Thanks.

Jongkeun: This event has not working like the `handleKeyDown`.
Because this is not key Event.
So I added it in a different way.

packages/selenium-ide/src/neo/components/TestRow/index.jsx
```diff
@@ -183,8 +183,18 @@ class TestRow extends React.Component {
   paste() {
     this.props.pasteFromClipboard(this.props.index);
   }
-  select() {
-    this.props.select(this.props.command);
+  select(event) {
+    if (event.ctrlKey || event.metaKey ) {
```
corevo: This is still flaky, `modifier-keys` is a library I wrote myself, so we can most certainly change it to fit the case.

packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -92,13 +93,19 @@ class UiState {
   }
 
   @action.bound copyToClipboard(item) {
-    this.clipboard = item;
+    this.clipboard.clear();
+    for(let i=0; i<this.selectedCommands.length; i++){
+      this.clipboard.push(this.selectedCommands[i]);
```
corevo: Why not
```js
this.clipboard.replace(this.selectedCommands)
```

packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -92,13 +93,19 @@ class UiState {
...
+    }
   }
 
   @action.bound pasteFromClipboard(index) {
-    if (this.clipboard && this.selectedTest.test) {
-      const newCommand = this.clipboard.clone();
-      this.selectedTest.test.insertCommandAt(newCommand, index);
+    if (this.clipboard.length && this.selectedTest.test) {
+      for(let i=0; i<this.clipboard.length; i++){
```
corevo:
```js
this.clipboard.forEach((command, clipboard, index) =>
```

packages/selenium-ide/src/neo/components/TestRow/index.jsx
Jongkeun: I tried to use `modifier-keys`, but although it works fine on a keyboard event,  not working on the mouse event.

Jongkeun:
@corevo Please checking this PR when you have a time.

packages/selenium-ide/src/neo/components/TestRow/index.jsx
corevo: Then we should change `modifier-keys` to work for mouse events as well.

Jongkeun: I couldn’t find the repository.
Is your private repository?

corevo: No, https://github.com/applitools/modifier-keys

Jongkeun: I tried to fix the project.
There is no part of a `shift` key and that library was not much different from this code.
Even you added the validation check for `MouseEvent`.
If I modify the repository, I think it has a long time.
Could you fix the library, please?

Jongkeun:
@corevo When will this be merged?

Jongkeun:
I can't reproduce this case.
Could you show me the error log at developer console?

corevo:
There's a bug in the dnd that I fixed at a later commit, if you can just pull those changes so that I can make sure multiple commands dnd works.  
Also when clicking a command, and then the one below it and clicking the arrow up key, the focus is lost, and arrow navigation is broken.

dvd900:
@Jongkeun Hey!! do you mind if pull code from this branch and try again?
 @corevo  any feed back for someone who's going to pursue this feature?

tourdedave:
If there is a problem, please open a new issue.

https://github.com/SeleniumHQ/selenium-ide/issues/new/choose

Jongkeun:
@dvd900 Sure. But it has a performance problem also. 
When there are 1000 commands, there is a slight delay.

PR #252, opened 2018-06-18
add Variables.js at the request of #220

Jongkeun:
add Variables.js at the request of #220


corevo:
I think you checked out the wrong [file](https://github.com/SeleniumHQ/selenium-ide/pull/252/files#diff-124a20c71367560dbc71391de4cc0b24).  
Also make sure to add it's tests, in #220 they are located in the `domain` folder, move them to the `view` directory instead.  
Also please squash this PR, because that mistake will make that commit fail.

Jongkeun:
Oh I see.
I don't know why there is in the `domain` folder.

corevo:
Yes, everything that isn't GUI, that will make formatCommand work with the new `Variables.js`

PR #254, opened 2018-06-18
Command Reference Tab

tourdedave:


packages/selenium-ide/src/neo/components/TabBar/style.css
```diff
@@ -34,3 +34,14 @@
...
   bottom: 0;
 }
+
+.tabbar .unread:after {
+  content: "";
+  display: inline-block;
+  height: 5px;
+  width: 5px;
+  margin: 2px 8px;
+  background-color: #EE4841;
```
corevo: I'll have to take a look at the styling with my eyes, but the correct color scheme to draw attention is `#40A6FF`, not your fault tho 😉

corevo:
Some points from eye checking.  
image
I don't like that the content can reach the edges, return the padding to all sides.  
Also something happened to the log overflow, fill the logs to the end, it doesn't overflow anymore.

packages/selenium-ide/src/neo/containers/Console/index.jsx
```diff
@@ -24,41 +24,53 @@ import { output } from "../../stores/view/Logs";
...
       <footer className="console" style={{
         height: this.props.height ? `${this.props.height}px` : "initial"
       }}>
-        <TabBar tabs={["Log", "Reference"]} tabWidth={90} buttonsMargin={0} tabChanged={this.tabChangedHandler}>
+        <TabBar tabs={tabs} tabWidth={90} buttonsMargin={0} tabChanged={this.tabChangedHandler}>
           <ClearButton onClick={output.clear} />
         </TabBar>
-        {this.state.tab === "Log" && <LogList output={output} /> }
-        {this.state.tab === "Reference" && <CommandReference /> }
+        <div className="viewport">
```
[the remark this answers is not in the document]
tourdedave: It looks like I could also just apply `overflow-y: auto` to `.viewport` instead of `.logs`.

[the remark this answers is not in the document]
corevo: Yes, but we are enforcing overflow on every tab.  
At the moment I don't see an issue with that.

packages/selenium-ide/src/neo/components/CommandReference/index.js
```diff
@@ -16,19 +16,47 @@
...
-export default class ReferenceMessage extends React.Component {
+export default class CommandReference extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
+    currentCommand: PropTypes.object
+  };
+  invalidCommand() {
+    return (
+      <li className="invalid-command">
+        <strong>Invalid command name provided.</strong>
```
corevo: Saying "Unknown command" is much better imo.  
Invalid is a bit negative, especially is it's just a typo.

packages/selenium-ide/src/neo/components/LogMessage/style.css
```diff
@@ -19,10 +19,6 @@
   color: #E80600;
 }
 
-.log .message:after {
-  padding: 0 5px;
```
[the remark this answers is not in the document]
tourdedave: I don't remember the specific reason. It might have been overzealous trimming. I'll look at putting it back.

packages/selenium-ide/src/neo/components/TabBar/style.css
tourdedave: :-)

packages/selenium-ide/src/neo/components/CommandReference/index.js
tourdedave: Okay.

packages/selenium-ide/src/plugin/manager.js
```diff
@@ -46,15 +46,26 @@ class PluginManager {
     });
   }
 
+  useExistingArgTypesIfProvided(target_type, docs) {
+    if (typeof(docs[target_type]) === "string" && docs[target_type] in ArgTypes) {
```
corevo: You are editing `docs` here, this argument is passed by ref and thus reassigning to it is suboptimal.  
You can use [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) to shallow copy the object.  
```js
return Object.assign({}, docs, {
  [target_type]: {
    name:....
  }
});
```

tourdedave: I was wondering about that. It didn't feel quite right mutating the original object.

PR #267, opened 2018-06-28
add a deprecated state

Jongkeun:

While I tested my old script, I saw the error count.
 
image

The test case had two errors which is the `isImplicitWait`.
But the test case was not failed.
Even the test case run correctly, I got the result failed because of `isImplicitWait`.

I thought the wait method is shown `failed` but not counted is better.


corevo:
Add before and after screenshots please

Jongkeun:
I think you don't need this PR.
I tested my branch which is removed `wait` filter.
So if I loaded the old script, there is the command related with `wait`.
I am going to apply to only my branch.
Sorry.

PR #280, opened 2018-07-03
WIP Control Flow v2

tourdedave:


packages/selenium-ide/src/neo/playback/playback-tree/index.js
```diff
@@ -0,0 +1,289 @@
...
+  [ControlFlowCommandNames.while]: trackControlFlowBranchOpen
+};
+
+function trackControlFlowBranchOpen (commandName, commandIndex, stack, state) {
+  state.push({ command: commandName, index: commandIndex });
+}
+
+function verifyElse (commandName, commandIndex, stack, state) {
+  if (!isIf(topOf(state))) {
+    throw "An else / else if used outside of an if block";
```
corevo: Always throw errors not strings, in this example I would throw [SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError)

packages/selenium-ide/src/neo/playback/playback-tree/index.js
```diff
@@ -0,0 +1,289 @@
...
+  [ControlFlowCommandNames.end]: levelBranchEnd,
+  [ControlFlowCommandNames.if]: levelBranchOpen,
+  [ControlFlowCommandNames.repeatIf]: levelBranchEnd,
+  [ControlFlowCommandNames.times]: levelBranchOpen,
+  [ControlFlowCommandNames.while]: levelBranchOpen
+};
+
+function levelDefault (command, level, levels) {
+  levels.push(level);
+  return level;
```
corevo: Avoid mutating `levels`, just like you return `level`
```js
return { level, levels: [...levels, level] }
```

packages/selenium-ide/src/content/selenium-api.js
```diff
@@ -228,8 +228,16 @@ Selenium.prototype.reset = function() {
...
+Selenium.prototype.eval = function(script, scoped = true, isExpression = false) {
+  if (isExpression) {
+    return window.eval(scoped ? `(() => (${script}))()` : script);
+  } else {
+    return window.eval(scoped ? `(() => {${script}})()` : script);
+  }
+};
+
+Selenium.prototype.doEvaluateConditional = function(condition) {
+  return !!(this.eval(condition, undefined, true));
```
corevo: Why not scoped?  
In WebDriver it'll be scoped.

tourdedave: I can make it explicit, but isn't it already scoped if it's `undefined`?

corevo: Forgot it has a default value `true`, make it true anyway for clarification that it is required.  
I put the default one, so that if it gets called from something internal that I'm not aware of, that it'll be scoped by default.

tourdedave: Okay.

PR #284, opened 2018-07-06
fixed error when migrating old test case which has a command

Jongkeun:

While I test an old test case, I got an error message.
The test case is a test case which has one command.
I don't know if you feel this is needed or not, I think this checking is not bad.

[oneCommand.zip](https://github.com/SeleniumHQ/selenium-ide/files/2168685/oneCommand.zip) is not the test case which I used, but I attach a test case which has one command.

And the `loadProject` function doesn't seem to work well about an old test case at the master branch.


corevo:
Add a test for this in `migrate.spec.js`

PR #293, opened 2018-07-11
set current tests before playing

Jongkeun:

Test step
1. I tested the "aa playback" test case.
2. After finishing, run another test case.
3. But the IDE is executing the previous script.

ezgif com-gif-maker


corevo:
That's not the correct fix, it has an opening for another issue, because `setCurrentTests` assumes the user of `changeView` knows `changeViews`' internal implementation.  
Fixed in 16f02fc9

PR #358, opened 2018-08-22
update docs about config file

M3ssman:
Minor documentation update for configuration file usage


corevo:
Keep the disabling behavior, some users unbeknownst to them may use a `.sideyml` when they don't want to.

M3ssman:
Current documented option states 

> You can disable that behavior with `--no-sideyml`.  

but that's not true, therefore the update. 
Since Release 3.2.0 (at least) you get an error ```error: unknown option `--no-sideyml'```.
So it's not possible to let side-runner ignore the file if it's named this way.

M3ssman:
Anything new regarding this topic? Or, it's considered a bug when one can't disable the default behavior with that flag?

corevo:
It's definitely a bug, and I should return that flag.

PR #421, opened 2018-09-27
Add wildcard support to assertText with the existing PatternMatcher

### LINKED ISSUE #141: Add support for "patterns" for text verification
Raising this after seeing another user talk about it.

The old IDE allowed regex in text verification steps (and possibly other places too) eg

verify text | target | you are the * User to log in today

This should be added to the new IDE as well to keep feature parity.

pekko:

Fixes #141.

Add wildcard/regex support by reusing existing `PatternMatcher` implementation.

Open question - should this be enabled also for other checks than text/notText?

### DIFF
--- a/packages/selenium-ide/src/content/selenium-api.js
+++ b/packages/selenium-ide/src/content/selenium-api.js
@@ -424,7 +424,7 @@ Selenium.prototype.findElementVisible = function(locator) {
 Selenium.prototype.doAssertText = function(locator, value) {
   const element = this.findElementVisible(locator);
   const visibleText = bot.dom.getVisibleText(element);
-  if (visibleText !== value) {
+  if (!PatternMatcher.matches(value, visibleText)) {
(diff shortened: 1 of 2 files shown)


corevo:
Yes it should, this should be implemented for every command that takes a [string pattern](https://github.com/SeleniumHQ/selenium-ide/blob/master/packages/selenium-ide/src/neo/models/Command.js#L188).  
Other than that this is also only a partial implementation, you also need to implement it for [selenium-side-runner](https://www.npmjs.com/package/selenium-side-runner), take a look [here](https://github.com/SeleniumHQ/selenium-ide/blob/master/packages/selianize/src/command.js).  
My intention was to implement this in the runner by converting globs to regexes, since JavaScript comes with a built-in regex engine, but not glob engine, this would let us not ship a glob implementation with the runner itself, but in Selianize instead.  
https://www.npmjs.com/package/glob-to-regexp

pekko:
ok - not as simple as I hoped, I'll come back with another PR later on

PR #485, opened 2018-11-19
"Play to this point" and "Record from here"

tourdedave:

### DIFF
--- a/packages/selenium-ide/scripts/jest/test.config.js
+++ b/packages/selenium-ide/scripts/jest/test.config.js
@@ -1,10 +1,10 @@
-import "jest-dom/extend-expect"; // matchers for view tests
+import 'jest-dom/extend-expect' // matchers for view tests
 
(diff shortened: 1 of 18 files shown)


Jongkeun:
@tourdedave , let me know why "play from here" changed to "play to this point"?
I have used the feature well.
I was embarrassed by the delete of that feature.

tourdedave:
Given the fact that playback has global state like control flow, variables, and callstack... you can no longer play from the middle of a test case. If you want to work with commands in the middle of a test case you can set a break point and play the test case -- it will pause at the break point in the debugger (this is what `Play to this point` does for you). You can then execute individual commands from the middle of the test case or resume the test case to play all of the remaining commands.

Jongkeun:
Yes, correct. I hope that my users are programmers. haha
They don't know why they have to classify the units independently.
Thank you.

Issue #510, opened 2018-12-02
Save scroll position separately for each test 


Yrds:
## 🚀 Feature Proposal

Save scroll position separately for each test 

## Motivation

This behavior could by handy, in cases editing two or more tests at same time. Sometimes it's annoying switch between tests and have to back to previously scroll position

## Example

On Test 1:
Scroll to top

On Test 2:
Scroll to bottom

Expected Result:

When back to Test 1, scroll position(top) should be on position i was before

## Screenshots

![captura de tela de 2018-12-03 11-55-53](https://user-images.githubusercontent.com/8484114/49377963-a248d200-f6f2-11e8-9a7d-d78c2ac7cf07.png)

Yrds:
Sorry to not specify. I added a screenshot highlighting the scrollbar

Yrds:
For sure, this would a leap of productivity

dvd900:
@corevo or anyone, have you started this? taking it now if not.

corevo:
@dvd900 great!

PR #520, opened 2018-12-11
Save last selected command per test to prevent the need to rescroll

dvd900:
I added the field selectedCommand to the TestState which I set in selectCommand. When a new test is selected, I check to see if that field is set and use if set.

Test pass the same as in Master, Lints with no errors. 

For the proposed feature:
https://github.com/SeleniumHQ/selenium-ide/issues/510


dvd900:
As a note: the reason why I had to do the scroll in test Row instead of test table was because the test row's mount occurs after the test table update.

dvd900:
Is there something more to be done for this?

corevo:
I'm on vacation this week, I'll review it next work week

corevo:
Thanks for contributing ❤️

Issue #503, opened 2018-11-26
Drop-down with alternative locators does not appear immediately after using "Select target in page"


barancev:
## 🐛 Bug Report

1) Insert a new command (say, click)
2) Use "Select target in page" button to "record" a locator
3) Try to see alternatives to the recorded locators -- fail, there is no dropdown.

To see the alternatives one should switch to another command and then back to the new one.

## To Reproduce

See the attached screencast:
[selenium_ide_no_alternatives.zip](https://github.com/SeleniumHQ/selenium-ide/files/2614269/selenium_ide_no_alternatives.zip)

## Expected behavior

Dropdown with alternative locators to appear immediately after "recording".

## Environment

OS: Windows 7
Selenium IDE Version:  3.4.4
Browser: Firefox
Browser Version: 65.0a1 (2018-11-25) (64-bit)

PR #525, opened 2018-12-23
Fix to show alternative locators immediately, rather than having trigger a re-render  

dvd900:
for: https://github.com/SeleniumHQ/selenium-ide/issues/503 


packages/selenium-ide/src/neo/IO/SideeX/find-select.js
```diff
@@ -104,8 +104,8 @@ function selectTarget(target) {
     UiState.selectedCommand.setTargets(target)
```
corevo: Let's add the fix here too

corevo:
Fixes #503  
Thanks! 👍

PR #529, opened 2019-01-07
changed from double keyDown to a keyDown and a keyUp for a KEY_ENTER

Jongkeun:
- [ X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

When you call `sendKeys`, `sendCommand` is called `keyDown` twice.
It looks working well, but I think this was your intention.


corevo:
It works well because most websites use `keyDown` to handle js, but this was definitely written in error, thanks for catching that!

PR #566, opened 2019-01-30
add some docs to selianize

Meir017:

let me know if this is the right direction or not


corevo:
Thanks for sending this, sorry it took me a bit to get back to you, we just have to decide which direction we want to go with, I'm thinking jsdoc is great.

Issue #544, opened 2019-01-22
Disabling a command means that it does not show the relevant reference section


p2635:
## 🐛 Bug Report

When a command is disabled, the usual reference part of the command is not shown in the bottom box.

## To Reproduce

1. Create any command
2. Ensure that the 'Reference' tab is selected
3. Disable the command (either by pressing the '//' button or prefixing the command with '//')
4. The Reference box will now say 'Unknown command name provided.'

![default](https://user-images.githubusercontent.com/4522927/51538039-1b89b900-1e48-11e9-9696-c12a6f5bd4ea.png)

## Expected behavior

The reference of the command should stay in my opinion.

## Project file reproducing this issue (highly encouraged)

I don't think this is required but let me know if you need one.

## Environment

OS: Windows 10
Selenium IDE Version:  3.4.5
Browser: Firefox
Browser Version: 64.0.2 (64 bit)

PR #582, opened 2019-02-21
If command description is available display in the refrence window even if disabled

dvd900:
For:
https://github.com/SeleniumHQ/selenium-ide/issues/544

- [X ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

### DIFF
--- a/packages/selenium-ide/src/neo/containers/Console/index.jsx
+++ b/packages/selenium-ide/src/neo/containers/Console/index.jsx
@@ -65,9 +65,14 @@ export default class Console extends React.Component {
     this.viewport.scrollTo(0, to)
   }
   render() {
-    const command = UiState.selectedCommand
-      ? Commands.list.get(UiState.selectedCommand.command)
-      : undefined
+    let command
+    if (UiState.selectedCommand) {
+      if (UiState.selectedCommand.enabled) {
+        command = Commands.list.get(UiState.selectedCommand.command)
+      } else {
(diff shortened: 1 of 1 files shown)


packages/selenium-ide/src/neo/containers/Console/index.jsx
```diff
@@ -65,9 +65,14 @@ export default class Console extends React.Component {
...
   render() {
-    const command = UiState.selectedCommand
-      ? Commands.list.get(UiState.selectedCommand.command)
-      : undefined
+    let command
+    if (UiState.selectedCommand) {
+      if (UiState.selectedCommand.enabled) {
+        command = Commands.list.get(UiState.selectedCommand.command)
+      } else {
+        command = Commands.list.get(UiState.selectedCommand.command.substr(2))
```
corevo: There is no need for the if the bug is not actually here.  
Look at [displayedName](https://github.com/SeleniumHQ/selenium-ide/blob/master/packages/selenium-ide/src/neo/models/Command/index.js#L63) the bug is that a command can be commented in 2 ways, either by using the button or manually prepending `//` to the command, which results in a different command name.

dvd900: looks like displayedName was working as intended, we just needed to use it :)

corevo:
Thanks 🐢

PR #596, opened 2019-03-04
checking isTerminal before executeCommand

Jongkeun:

gif

Test case : [selectWindow.txt](https://github.com/SeleniumHQ/selenium-ide/files/2924966/selectWindow.txt)

I found out some bug.
If I disable a command that be set `New Window configuration`, it returns an error.
Because of `afterCommand` method in `ext-command.js`.
The waitForNewWindow is set true by `beforeCommand` and it is executed `waitForNewWindowToAppear`.

But there is no new window because we disabled the command.
Then an error will occur.

So I changed a location of `isTerminal`.
If the command was disabled, not execute `beforeCommand` and `afterCommand`.
I think this is right.

### DIFF
--- a/packages/selenium-ide/src/neo/playback/playback-tree/command-node.js
+++ b/packages/selenium-ide/src/neo/playback/playback-tree/command-node.js
@@ -67,6 +67,11 @@ export class CommandNode {
           'Max retry limit exceeded. To override it, specify a new limit in the value input field.',
       })
     }
+    if (this.isTerminal()) {
+      return Promise.resolve(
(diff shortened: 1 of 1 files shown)


corevo:
I think the correct fix is to not reuse `itTerminal` for both cases, `isTerminal` should check if the command is a terminating control flow command, and on top of that we should have a check `isDisabled`.  
And that should be used in `execute` prior to `beforeCommand`.

corevo:
Thanks 🐢

Issue #607, opened 2019-03-08
Save dialog broken once a Selenium IDE plugin is registered


zewa666:
## 🐛 Bug Report / Documentation request

When registering a custom plugin, the save button does not anymore open up a Dialog.
I've worked around the issue by listening to the `emit` action and sending a `true` response

```typescript
chrome.runtime.onMessageExternal.addListener((message, sender, sendResponse) => {
  // tslint:disable-next-line:no-console
  console.log("Event from IDE:", message);
  if (message.action === "emit") {
    sendResponse(true);
  }
});
```

## To Reproduce

* Use a minimal Extension sample
* Startup Selenium IDE
* Register with the plugin
* Click on the save 

## Expected behavior

The save dialog should popup without having to respond. Alternatively it'd be nice to have some documentation that tells that this has to be done and about what we're supposed to do here. Is this meant as a check for the plugin to prohibit saving unstable projects?

Moreover the action emit and entity project is barely understandable and might conflict with other calls, so how is the consumer supposed to know the event is about saving the project?

## Environment

OS: MacOS
Selenium IDE Version: 3.5.8  
Browser: Chrome
Browser Version: 72

corevo:
It is a bug with Chrome that I've filed with them [here](https://bugs.chromium.org/p/chromium/issues/detail?id=922373), due to this bug you are required to always answer the `browser.runtime.onExternalMessage` or the Promise won't resolve in the IDE side.  
Once Chrome fixes it you won't have to have this workaround in place.

zewa666:
Where can I update the docs to add the workaround info? It seems the repositories docs aren't in sync with the website.

tourdedave:
You could add it to the FAQ -- https://github.com/SeleniumHQ/selenium-ide/blob/master/docs/introduction/faq.md

PR #614, opened 2019-03-12
feat(plugin): send save instead of emit for save action

zewa666:
related issue  https://github.com/SeleniumHQ/selenium-ide/issues/607

### DIFF
--- a/packages/selenium-ide/src/neo/IO/filesystem.js
+++ b/packages/selenium-ide/src/neo/IO/filesystem.js
@@ -104,7 +104,7 @@ function downloadProject(project) {
 }
 
 function exportProject(project) {
-  return Manager.validatePluginExport(project).then(() => {
+  return Manager.validatePluginExport(project, 'save').then(() => {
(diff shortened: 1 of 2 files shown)


zewa666:
@corevo and @tourdedave as discussed in the related topic, how about we send a specialized emit action type called `save`, so that plugins can handle this case accordingly. Moreover this would simplify the documentation of the workaround needed for the current Chrome issue

corevo:
This is a breaking change for the API, and unless we draft API v2, I don't think we can change it at this point.  
This will probably have to be fixed through documentation instead.

zewa666:
So I've looked through the source and it seems like the combination of `emit` and `project` is used only for the save dialog scenario. Is that right? If so, yeah that can be handled via documentation only. Waiting for your reply then I'll create the docs PR

zewa666:
closing in favour of https://github.com/SeleniumHQ/selenium-ide/pull/617

Issue #537, opened 2019-01-13
Select Frame command not working


pdxwaitlist:
## 🐛 Bug Report

I am unable to use the selectFrame command to interact with elements inside an iframe on an HTML page

## To Reproduce

Use the following Selenium IDE commands to try to click the checkbox on a Google Recaptcha v2, which uses iframes to display on a page:
    {
      "Command": "open",
      "Target": "https://www.google.com/recaptcha/api2/demo",
      "Value": ""
    },
    {
      "Command": "selectFrame",
      "Target": "index=0",
      "Value": ""
    },
    {
      "Command": "click",
      "Target": "//*[@id=\"recaptcha-anchor\"]/div[5]",
      "Value": ""
    }

## Expected behavior

Actual result: (from Selenium IDE logs)
Running 'Untitled' 
1.open on google.com/recaptcha/api2/demo... OK 
2.selectFrame on index=0... Failed: Timeout

This also happens if using other methods to try to select the frame - by ID, by other indexes, etc.

Expected result: no error when running the selectFrame command

## Environment

OS: Mac OSX 10.13.6
Selenium IDE Version:  3.4.5 
Browser: Chrome
Browser Version: 71.0.3578.98

tourdedave:
Only index based locators for `select frame` are available right now.

Unfortunately I'm unable to reproduce the issue with the use case you provided.

Can you attach your project file to the issue as a zip file?

pdxwaitlist:
I did a little more troubleshooting on the issue. Strangely, this seems to be specific to Chrome -- when I run the test in Firefox it works normally but in Chrome the `select frame` command (using an index locator) returns an error. 

I've attached a zip file with my test case, as well as screenshots of the results when I run this test case in Chrome and in Firefox. Let me know if there is any other info I can provide. Thanks!

[iframe.zip](https://github.com/SeleniumHQ/selenium-ide/files/2771634/iframe.zip)

p2635:
Windows 10
Selenium 3.4.5
Firefox 64.0.2 (64bit)
Chrome Version 71.0.3578.98 (Official Build) (64-bit)

Tested both the original test and the zip file, it works for me but I am on Windows.

Jongkeun:
@tourdedave , Are there any plans that make available to find frame by `id` or stuff ? not only index.

corevo:
[...]

Issue #423, opened 2018-09-27
Issue locating/using elements within an iFrame using SelIDE


98gmarquee:
## 🐛 Bug Report

I'm unable to focus or interact with buttons/links/elements in an iframe via ff quantum or chrome using the current IDE.

## To Reproduce
Access a page that has buttons/links inside an iframe without an id, only src attribute and attempt to click these buttons. It's unable to find them. 

Also try using select frame using xpath to locate, such as
`select frame| xpath=//iframe[@src='/foo/foo.jsp'| `

Also tried recording and it recorded a command such as:
`select frame|Index=0 | `
Then when I try running it, it gets stuck on that line and throws an error.

It cannot find the frame to then focus on to run commands in that frame.

## Expected behavior

It should be able to find the elements within the frame.

## Project file reproducing this issue (highly encouraged)
I cannot produce the side file as it's accessing an internal web site. I've attached an image of the source for the iFrame including the button I'm attempting to locate/click highlighted. Hopefully that is sufficient.
![selide-iframeissue](https://user-images.githubusercontent.com/33339926/46160417-4c663280-c248-11e8-888a-67711227bbcb.png)

That is of this button:
![selide-iframeissue-screen](https://user-images.githubusercontent.com/33339926/46161439-2b531100-c24b-11e8-9fc7-15c52418920d.png)

## Environment
OS: w10
Selenium IDE Version:  3.3.1
Selenium SIDE Runner Version:  n/a
Node version: n/a
Browser:
Browser Version: firefox 62.0.2 64bit, chrome  69.0.3497.100 64bit

luquitia:
I'm also having this issue. I'm not sure how to structure the argument for **select frame** command; the reference tab is vague on this point (and the mentioned index options don't work for me).

I have an iframe with attributes like so: <iframe scrolling="no" allowtransparency="true" id="profileBuilder" title="Create Your Profile" style="border: 0px none; height: 297px;" src="/LongURL-thatIremovedHere" tabindex="0">

I tried:
select frame | id=profileBuilder
select frame | index=0
select frame | index=1
select frame | xpath=//iframe[@id='profileBuilder']
and some other variations
-- all throw errors like:  Failed: Invalid argument or Failed: Timeout

_Note: recording doesn't work for me as the playback doesn't recognize clicks within this iframe (since I guess it is not "selected")._

jngale:
[...]

PR #621, opened 2019-03-15
moved selectFrame command to brower side.

Jongkeun:

I tried to change a `selectFrame` command. 
Because I need to execute like old version. ( #537, #423)
If we do like this, we don't have to manage `tabId` anymore.
When I call a `selectFrame` command, `this.currentWindow` is set the frame that you selected.

Please try to check out this way.

test : [selectFrame.txt](https://github.com/SeleniumHQ/selenium-ide/files/2969948/selectFrame.txt)


corevo:
This won't work for CORS frames

Jongkeun:
Yes, you were right.
I hope that I can find frames by `id` soon.

PR #625, opened 2019-03-19
selianize support parse execute script with literal argument

bolasblack:
I need execute script with some literal argument (Document https://screener.io/docs/webdriver#/snapshots)

### DIFF
--- a/packages/selianize/__tests__/command.spec.js
+++ b/packages/selianize/__tests__/command.spec.js
@@ -281,40 +281,92 @@ describe('command code emitter', () => {
     )
   })
   it('should emit `run script` command', () => {
-    const command = {
+    const standardCommand = {
       command: 'runScript',
       target: "alert('test');\nalert('Im annoying');",
       value: '',
     }
-    return expect(CommandEmitter.emit(command)).resolves.toBe(
-      `await driver.executeScript(\`${command.target}\`);`
-    )
+    const valuedCommand = {
+      command: 'runScript',
+      target: "alert('test');\nalert('Im annoying');",
+      value: '0, "hello"',
+    }
+    return Promise.all([
+      expect(CommandEmitter.emit(standardCommand)).resolves.toBe(
+        `await driver.executeScript(\`${standardCommand.target}\`);`
+      ),
+      expect(CommandEmitter.emit(valuedCommand)).resolves.toBe(
+        `await driver.executeScript(\`${valuedCommand.target}\`,${
+          valuedCommand.value
+        });`
+      ),
+    ])
   })
   it('should emit `execute script` command', () => {
-    const command = {
+    const standardCommand = {
+      command: 'executeScript',
+      target: 'javascript',
+      value: '',
+    }
+    const varCommand = {
       command: 'executeScript',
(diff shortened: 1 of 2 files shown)


corevo:
This is a change to the specifications of `run script` `execute script` and `execute async script`.  
To change a command from its existing implementation, it requires documentation changes, IDE changes, and migration in case the change is backwards in compatible.  

But before making such changes to commands, we require explanation of why changes like these are required to the benefit of our users.

bolasblack:
@corevo You are right, there are some other things to do.

I this PR is just a proposal, I want to know what you think.

> we require explanation of why changes like these are required to the benefit of our users.

To be honest, I don’t know how to say. At present, this kind of scene should be very rare. It may be that only this kind of scenario is almost necessary for me. Other scenarios have almost any other solution 😂.

> it requires documentation changes
> IDE changes

I can do it

> and migration in case the change is backwards incompatible

I have no idea about how to keep compatibility 🤦‍♂️, or maybe I can change a syntax?

corevo:
Why can't variables fit? or hardcode it, since it is the value field anyway?

bolasblack:
@corevo If I config command like this:

```json
    {
      "id": "17bc0b97-35cc-4066-b169-311ba35f9c15",
      "comment": "",
      "command": "runScript",
      "target": "/*@screener.snapshot*/, 'afterLogout'",
      "targets": [],
      "value": ""
    }
```

the emitted is like this:

```js
await driver.executeScript(`/*@screener.snapshot*/, 'afterLogout'`);
```

But I hope I can emit code like this:

```js
await driver.executeScript(`/*@screener.snapshot*/`, 'afterLogout');
```

I tried:

```json
    [{
      "id": "88db4bb7-e3a2-46ba-9c35-ff1fcc8f0156",
      "comment": "",
      "command": "store",
      "target": "\"afterLogout\"",
      "targets": [],
      "value": "Screen State"
    }, {
      "id": "17bc0b97-35cc-4066-b169-311ba35f9c15",
      "comment": "",
      "command": "runScript",
      "target": "/*@screener.snapshot*/, ${Screen State}",
      "targets": [],
      "value": ""
    }]
```

But the emitted code

```js
vars["Screen State"] = `"afterLogout"`;
await driver.executeScript(`/*@screener.snapshot*/, arguments[0]`, vars["Screen State"]);
```

It looks like there are vary strict

I really don't know how to do it, so I think maybe I can create a PR to make selenium-ide support it.

corevo:
I understand the use case now, in this case I would recommend building a plugin for the IDE.  

When plugins introduce new commands, they can also emit code as part of Selianize, so you could implement the `driver.executeScript` that they want.

The API may look like `screener snapshot | snapshot name` and Selianize would know to convert it to code, instead of overloading `execute script`.

Take a look at the [plugin API docs](https://www.seleniumhq.org/selenium-ide/docs/en/plugins/emitting-code/)

bolasblack:
@corevo I knew I can implement this feautre by create a plugin, but I thought it is too simple and maybe can be an "improvement" of Selenium IDE

But maybe you are right, I should probably use a plugin to implement this function.

PR #644, opened 2019-04-05
Multi select

dvd900:
Would be happy to port this to the electron app as well. 

Credit to @Jongkeun for initial legwork. 
 
- [ X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

### DIFF
--- a/packages/selenium-ide/src/neo/components/TestRow/index.jsx
+++ b/packages/selenium-ide/src/neo/components/TestRow/index.jsx
@@ -108,6 +108,7 @@ class TestRow extends React.Component {
   static propTypes = {
     index: PropTypes.number,
     selected: PropTypes.bool,
+    lastSelected: PropTypes.bool,
     className: PropTypes.string,
     status: PropTypes.string,
     readOnly: PropTypes.bool,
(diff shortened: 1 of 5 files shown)


corevo (changes requested):
We are using `yarn` how is it that a `package-lock.json` was generated, please remove it and make sure to squash so it won't be in the history.

dvd900:
@corevo oh, weird. Squashed it down to one commit for easier viewing.

corevo (changes requested):
Hey, great work but there are a few experience improvements that are needed  
- drag n drop for the entire selection
- holding down the primary modifier (ctrl for windows command for mac) should be additive to the selected commands, currently it has no effect (look at modifier keys to see how we handle it cross os)
- only the last selected command should have the top border (it is used to denote where commands will be inserted when recording)

Thanks a lot for the work put into this, this looks great so far!

tourdedave:
@dvd900 Hey, have you had a chance to look at @corevo's improvement requirements to land this PR? I can lend a hand if you'd like.

Issue #618, opened 2019-03-13
Set the value to a default when using a "wait for" command


98gmarquee:
## 🚀 Feature Proposal

When using commands that require a wait time such as ones that start with "wait for", it would be nice if it set the value to a default value like 30000 (30 seconds).

## Motivation

Helps make writing tests quicker

## Example

Select command "wait for element not visible", set the wait time to 30000 automatically.

corevo:
That's a great idea, sure!

98gmarquee:
I think I saw the same thing happen when importing old IDE scripts and it would change whatever the old command was to the new command, but leave the field blank, so it would fail when you ran it. I couldn't find my example of that, so I didn't include it. Just thought I'd mention it here in case someone gets a chance to look at it.

98gmarquee:
Just updating for a reminder about the setting of a default wait when importing an old ide script.

harsha509:
Closing as resolved!

PR #678, opened 2019-05-08
Added set default value when select 'wait for...' command

sapzape:
Set the default value 3000 (30 seconds) when select 'wait for...' command.
image

Related issue #618

### DIFF
--- a/packages/selenium-ide/src/neo/__test__/models/Command.spec.js
+++ b/packages/selenium-ide/src/neo/__test__/models/Command.spec.js
@@ -207,17 +207,18 @@ describe('Command', () => {
     expect(command instanceof Command).toBeTruthy()
   })
 
-  it('should set default value in the wait for command when there is no value', () => {
+  it('should not overridden value when replace command', () => {
     const command = new Command()
+    command.setCommand('assert text')
+    command.setValue('Hello World')
     command.setCommand('wait for element present')
-    expect(command.value).toBe('30000')
+    expect(command.value).toBe('Hello World')
   })
 
-  it('should set value in the wait for command when there is a value', () => {
+  it("should not set value for commands that do not start with 'wait for'", () => {
     const command = new Command()
-    command.setCommand('wait for element editable')
-    command.setValue('50000')
-    expect(command.value).toBe('50000')
+    command.setCommand('click')
+    expect(command.value).toBe('')
   })
 })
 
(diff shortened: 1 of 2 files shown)


packages/selenium-ide/src/neo/models/Command/index.js
```diff
@@ -106,6 +106,10 @@ export default class Command {
     if (!this.canHaveTargets) {
       this.setTargets()
     }
+
+    if (this.command.indexOf('waitFor') > -1) {
```
corevo: This could potentially override a command that has been changed to wait, or override a current wait command with a different value, check that value is empty string.

sapzape: fixed it.

sapzape:
I just used two 'wait for' commands as tests to check the value setting.
Is it the correct test code?

packages/selenium-ide/src/neo/__test__/models/Command.spec.js
```diff
@@ -206,6 +206,19 @@ describe('Command', () => {
...
+
+  it('should set default value in the wait for command when there is no value', () => {
+    const command = new Command()
+    command.setCommand('wait for element present')
+    expect(command.value).toBe('30000')
+  })
+
+  it('should set value in the wait for command when there is a value', () => {
+    const command = new Command()
+    command.setCommand('wait for element editable')
```
corevo: Change this test to check the following flow
 - created a store command with some value
 - replaced it with a `wait for` command
 - `value` did not get overridden

packages/selenium-ide/src/neo/__test__/models/Command.spec.js
```diff
@@ -206,6 +206,19 @@ describe('Command', () => {
...
+    command.setValue('50000')
+    expect(command.value).toBe('50000')
+  })
 })
```
corevo: Add another test that checks that the value does not get set for commands that do not start with `wait for`

packages/selenium-ide/src/neo/__test__/models/Command.spec.js
sapzape: Fixed it.

packages/selenium-ide/src/neo/__test__/models/Command.spec.js
sapzape: Fixed it.

sapzape:
PTAL :)

98gmarquee:
Glad to see this didn't get forgotten. It keeps hanging me up! lol

Issue #552, opened 2019-01-23
Log output > Include timestamp against each command


p2635:
## 🚀 Feature Proposal

When running tests, it would be useful if there was an option to include the timestamp (and date?) in the log output. For example [01/01/2019 13:49] 1. click on bob OK.

## Motivation

- Useful for debugging
- Useful for checking how long it took for particular parts of the test

## Example

Perhaps we can add a settings section and in there, the user can select whether or not they want the date/time to be included.

corevo:
Is there a standard we can follow, there is a probable need for localization.

PR #688, opened 2019-05-15
 Added timestamp against each log

sapzape:
This is an implementation to provide timestamp to client selenium ide log.
'toLocaleString ()' returns the year, month, and day of the month in a time format appropriate to each culture.

include_timestamp_log

Related issue #552

### DIFF
--- a/packages/selenium-ide/src/neo/components/LogMessage/index.jsx
+++ b/packages/selenium-ide/src/neo/components/LogMessage/index.jsx
@@ -37,6 +37,12 @@ export default class LogMessage extends React.Component {
         statusMessage = 'Undetermined'
       }
     }
+
+    let hours = ("0" + this.props.log.timestamp.getHours()).slice(-2)
+    let minutes = ("0" + this.props.log.timestamp.getMinutes()).slice(-2)
+    let seconds = ("0" + this.props.log.timestamp.getSeconds()).slice(-2)
+    let time = hours + ":" + minutes + ":" + seconds
+
     return (
       <li
         className={classNames('log', this.props.log.status, {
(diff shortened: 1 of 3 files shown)


packages/selenium-ide/src/neo/ui-models/Log.js
```diff
@@ -34,10 +34,13 @@ export default class Log {
   channel = null
   @observable
   isNotice = false
+  @observable
```
tourdedave: A log is a single entry in the console, so there's no need to track it in an observable.

packages/selenium-ide/src/neo/ui-models/Log.js
```diff
@@ -34,10 +34,13 @@ export default class Log {
...
+  time = false
 
   constructor(message, status) {
     this.message = message
     this.status = status
+    this.setTime()
```
tourdedave: To simplify this, it would be better to just do the timestamp creation here in the constructor and handle formatting in the UI (removing the need for a separate bounded function).

Also, a better name would be `timestamp`. So, something like `this.timestamp = new Date()`.

packages/selenium-ide/src/neo/components/LogMessage/index.jsx
```diff
@@ -49,6 +49,7 @@ export default class LogMessage extends React.Component {
           }}
         >
           <div className="log-overview">
+            <span className="time">[{this.props.log.time}]</span>
```
tourdedave: re: formatting in the UI -- it would be good to just show the time w/ seconds (no date).

packages/selenium-ide/src/neo/ui-models/Log.js
sapzape: Fixed it.

packages/selenium-ide/src/neo/ui-models/Log.js
sapzape: Fixed it.

packages/selenium-ide/src/neo/components/LogMessage/index.jsx
sapzape: Fixed it.

sapzape:
PTAL :)

tourdedave:
Thanks!

Issue #263, opened 2018-06-25
Chrome Extension stuck in infinite loop on site


ChadKillingsworth:
When loading a large single page app, this extension gets stuck in an infinite loop and locks up the tab. It takes progressively more and more memory. I've verified that this occurs with the Chrome extension on both Windows and Mac.

Of note: the site in question uses Web Components and Shadow DOM. I do not know if that effects the issue or not.

By pausing the script execution in the dev tools, I get the following stack trace. I can infinitely step through the top `getframeLocation` call.

```
getframeLocation (injector.js:42102)
./content/record-api.js (injector.js:42102)
__webpack_require__ (injector.js:30)
./content/prompt-injector.js (injector.js:42059)
__webpack_require__ (injector.js:30)
1 (injector.js:50730)
__webpack_require__ (injector.js:30)
(anonymous) (injector.js:73)
(anonymous) (injector.js:76)
webpackUniversalModuleDefinition (injector.js:7)
(anonymous) (injector.js:10)
```

corevo:
What might affect this is frames, can you tell me the frame structure in your app?

ChadKillingsworth:
The main page where this is happening uses [ChartJS](https://www.chartjs.org/). It draws on a canvas in an iframe.

<img width="864" alt="screen shot 2018-06-26 at 6 46 54 am" src="https://user-images.githubusercontent.com/1247639/41909551-e464e6a6-790c-11e8-9b84-acceba74774d.png">

It's not nested inside ShadowDom in this case, but is the light dom child of a custom element.

corevo:
Shadow DOM is highly unlikely to affect this, that function mostly looks for `iframes`.  
Thanks for the info, I'll try to reproduce this.

corevo:
It's not ChartJS inside an `iframe` alone, perhaps something is nested inside the `iframe`?  
Can you give me a better picture of the DOM?

corevo:
Can you remove remove the charts and confirm for me wether the issue persists?

Cobaltikus:
[...]

PR #739, opened 2019-07-05
Issue #263 Chrome Extension stuck in infinite loop on site

Cobaltikus:
window.parent.frames does not include iframes added dynamically, causing an infinite loop - so use getElementsByTagName instead


Cobaltikus:
A link to the issue this fixes:
https://github.com/SeleniumHQ/selenium-ide/issues/263

Cobaltikus:
window.parent.frames DOES include iframes added dynamically

The real issue was a bad line of js clobbering length = 0, which messes up window.length and window.frames.length.

corevo:
Thank you for following up on the pr.

Cheers,
Tomer

PR #752, opened 2019-07-24
Bug fix: Foreach command in selenium-side-runner

cdifino:
The selenium side runner had an off by one bug in the foreach command. When looping through an array it would miss the last element of the array.

The underlying problem of this error was found in selianize/src/command.js file, where the loop of the array was written "for (let i = 0; i < vars["collection"].length -1; i++)" instead of "for (let i = 0; i < vars["collection"].length; i++)". This caused the program to not loop through the last element of the array when running the foreach command with the selenium-side-runner.


cdifino:
Hi @tourdedave ! This is the last and final PR for the small bug. Thank you in advance for reviewing!

tourdedave:
Thanks for the PR! I'll review soon.

Issue #765, opened 2019-08-02
Accessing Object Values


cdifino:
## 🚀 Feature Proposal

To access a property value of a stored json object, I need to do a workaround using execute script command each time(ex.  "execute script, return ${object}.someKey, someVariable"). This workaround  causes my .side scripts to look very messy and longer than needed. 

Instead of adding a new command just to access the property, I would like to do this: ${object.someKey}" any time I need to use that property value. The idea would be to modify the xlateArgument function so if the variable "object.someKey" is not found, parse it to see if "object" is found to be a variable and then access the object property value and return that as the argument. We would also have to evaluate the changes for supporting the code export of this feature. 

## Motivation

In Selenium IDE I need to access object properties previously stored using the storejson command, or provided as inputs to a Selianize test. Currently, accessing a property of an object requires a separate command step to execute JavaScript, which is not intuitive. I can currently access properties of objects directly when using Selianize using the proposed syntax ${objectVar.property}, and the extension and other runners should support this as well.

## Example

if we want to implement the following script it would be very beneficial and it would save unnecessary commands to extract property values. 

### **How it currently works:**

command: store json 
target: { "People" : [{"name":"John", "age":30, "car":null }, {"name":"Paul", "age":27, "car":null }]} 
value: json

**command: execute script
target: return ${json}.People
value:people**

command: for each 
target: people 
value: i

**command: execute script
target: return ${i}.name
value: name**

command: type 
target: name=q
value: ${name}

command: end
target: 
value:

### **Our proposal:** 

command: store json 
target: { "People" : [{"name":"John", "age":30, "car":null }, {"name":"Paul", "age":27, "car":null }]} 
value: json

command: for each 
target: json.People
value: i

command: type 
target: name=q
value: ${i}.name

command: end
target: 
value:

tourdedave:
@cdifino PR welcome :-)

corevo:
He can't send that PR, we need to spec this behavior, how would we know in string interpolation if ${a}.b should be $$$.b or the b property on a?

That's why I was thinking we should move to ${a.b} but that's complex.

Cheers,
Tomer

cdifino:
Hi,

 I was thinking on accessing properties doing ${a.b}. Is it ok if I raise a
PR that works with that approach?

Thanks,

Camila

tourdedave:
Thanks for the PR! We'll review it (#768) soon.

billxiao:
[...]

PR #766, opened 2019-08-06
Accessing Object Values #765 

cdifino:


cdifino:
Hello @corevo and @tourdedave! 

This is the basic concept of what I think this implementation would look like. As of now, this PR implemented the following feature and fix a bug along the way. 
- It is able to access property values of json objects  (ex. ${a.b.c}, ${a.b[0].c} )
- It fixed the bug that when the array variable in foreach is undefined, it would get in an infinite loop instead of throwing the "Invalid variable provided."

This feature/PR is not yet complete, I am still missing the following: 
- Testing property access of json
- Applying the feature to for each and verify (Only two commands that read variables without brackets) 
- Exporting feature to other languages

I am looking for a good solution to apply the feature to 'for each' and 'verify'.  Currently evaluateForEach() formats the command using xlateArgument (This function process ${} as variable and everything else as simple string). 
For for each case, the best solution I could propose would be to create a separate function inside "formatCommand.js" to process the command's target as a variable without curly brackets instead of a simple string. 
Another solution could be to finish processing the command in the function "evaluateForEach" inside "command-node.js". 

Thanks,

corevo:
Thanks for helping out 🥂 

Another thing that I'm a bit concerned about is `master`, we can't afford the electron rework to regress on the extension, so the fix will have to be forward ported.

PR #768, opened 2019-08-08
Object Property Access

cdifino:
- [x ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)

PR #797, opened 2019-09-06
API Extension to open SeleniumIDE from external event

cdifino:


cdifino:
Hi @corevo! 

This PR contains the implementation of an API method that will launch SeleniumIDE Extension if it is not open. If the call contains a project, it will load it, if not, it will just open the extension. 

You can give it a try by installing the extension that I will send to you ( I could not attach the extension files to this comment) . When the extension is installed, when clicking on the extension icon, this will make a call to the API. The payload of the API call is where the desired project to load will be placed. You can try it out and let me know what you think. 

Regards, 

Camila

corevo:
This PR contains #768, so we have two options to merge it, either we fix #768 and then we can merge this after this is rebased to master, or rebase it now and remove the dependency on one another.

cdifino:
Sorry, will remove the dependency. So you want to rebase this to the master/electron branch?

PR #804, opened 2019-09-09
Extending api to open SeleniumIDE and load projects

cdifino:


cdifino:
@corevo  this is detached to the other PR

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -60,6 +62,14 @@ router.get('/project', (_req, res) => {
   res({ id: UiState.project.id, name: UiState.project.name })
 })
 
+router.post('/project', (_req, res) => {
+  if (_req.id) {
```
corevo: Should be `req.project.id` so we can maybe add more metadata if necessary in the future. (also no `_` required)

packages/selenium-ide/src/background/background.js
```diff
@@ -191,12 +191,22 @@ browser.runtime.onMessageExternal.addListener(
       .sendMessage(message)
       .then(sendResponse)
       .catch(() => {
-        return sendResponse({ error: 'Selenium IDE is not active' })
+        if (message.verb == 'post' && message.uri == '/project') {
+          initiateWindow(function() {
+            browser.runtime.sendMessage(message)
```
corevo: Don't combine the two API calls, as far as I see it there are 2 calls:
 - Open Selenium IDE
 - Load a project

So this one should only load it, and return true if it was loaded, also since the endpoint doesn't exist the payload should be different as well.
```js
{
  openSeleniumIDE: true
}
```

Instead of the usual `{verb, uri, message}`.

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -60,6 +62,14 @@ router.get('/project', (_req, res) => {
...
+    loadJSProject(UiState.project, _req)
```
corevo: If the current project is [modified and not saved](https://github.com/SeleniumHQ/selenium-ide/blob/v3/packages/selenium-ide/src/neo/stores/domain/ProjectStore.js#L29), we should show them a dialog confirming that they are about to lose data.

Mentioning that `[plugin name] is trying to load a project, unsaved changes will be lost, are you sure?`

corevo:
Merged in a7a4d358f294cae53878ef89b4fb93b5ff425000

PR #813, opened 2019-09-16
Api plugin registration verification

cdifino:


corevo:
I reverted this changeset because it has caused multiple plugins to stop functioning.

I'm still not sure what to do about this, I think we can decorate new APIs to start, and figure out a date when we could make this change happen API-wide.

cdifino:
ok, that is interesting. Can I still protect the load method agains this issue? For the moments? @corevo

PR #823, opened 2019-09-21
Connected Mode

cdifino:
- [x ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/selenium/blob/master/CONTRIBUTING.md#step-6-sign-the-cla)


cdifino:
Hello @corevo  here is the modifications I have done for the connected mode feature. 
Here is how it currently works: 
If it is in connected mode and we want to interact with Selenium IDE, it will always need to match the connection id. If it is not in connected mode, there is no need to pass a connection ID in the payload and everything should work fine. 
The only way to create another connection is by calling "connect with the new connection id"
When connecting to a new extension it will prompt the user is he/she wants to connect. 
If extension wants to load project to Selenium IDE, and the user has unsaved changes, it will prompt the user to save files or just load. This will happen if there is a connection or not. 
If you need an extension to test these changes I made this little extension. 
[ext.zip](https://github.com/SeleniumHQ/selenium-ide/files/3639021/ext.zip)

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -59,40 +63,54 @@ router.post('/log', (req, res) => {
...
+  if (_req.connectionId != UiState.isControlledBy.connectionId)
+    throw new Error('Another connection is established')
   res({ id: UiState.project.id, name: UiState.project.name })
 })
 
+router.post('/connect', (req, res) => {
+  if (UiState.isControlledBy.connectionId != req.connectionId) {
+    if (!ModalState.welcomeState.completed) ModalState.hideWelcome()
+    ModalState.showAlert({
+      title: 'Connect',
```
corevo: `Connect` is a confusing term, I think `Control` is a much better one.

I think the title should be `Assisted Control` and body `[plugin name] is trying to control Selenium IDE`, with labels being `Restart and allow` and `Deny access`

packages/selenium-ide/src/background/background.js
```diff
@@ -191,9 +191,9 @@ browser.runtime.onMessageExternal.addListener(
       .sendMessage(message)
       .then(sendResponse)
       .catch(() => {
-        if (message.openSeleniumIDEIfClosed) {
+        if (message.uri == '/connect' && message.verb == 'post') {
```
corevo: No reason not to keep them both

corevo: especially since its in production and documented.

packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -67,6 +67,12 @@ class UiState {
   options = {}
   @observable
   pauseNotificationSent = false
+  @observable
+  isControlledBy = {
```
corevo: This state belongs in the PluginManager, just add a boolean so you can render the top banner.

packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -67,6 +67,12 @@ class UiState {
...
+    connectionId: null,
```
corevo: Not sure why we need `connectionId` we can use the `extensionId` instead.

packages/selenium-ide/src/neo/stores/view/UiState.js
corevo: I was thinking `PluginManager.controller` with `controller` being the plugin instance.

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -34,6 +34,8 @@ router.get('/health', (req, res) => {
 })
 
 router.post('/register', (req, res) => {
+  if (req.connectionId != UiState.isControlledBy.connectionId)
+    throw new Error('Another connection is established')
```
corevo: I think we can register other plugins, we should just block requests to `/control` so only one can control it, thoughts?

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -59,40 +63,54 @@ router.post('/log', (req, res) => {
 })
 
 router.get('/project', (_req, res) => {
+  if (_req.connectionId != UiState.isControlledBy.connectionId)
+    throw new Error('Another connection is established')
   res({ id: UiState.project.id, name: UiState.project.name })
 })
 
+router.post('/connect', (req, res) => {
```
corevo: `/control`

corevo: The better way to do this is via a closure as you can re-use it
```js
const controlledOnly (apiFn) => {
  return (req, res) => {
    if (req.connectionId != UiState.isControlledBy.connectionId)
      throw new Error('Selenium IDE is controlled by a different extension')
    return (apiFn(req, res))
  }
}
```

And then you can use it like so
```js
router.get('/control', controlledOnly((req, res) => {}))
```

corevo: We'll need 2 closures, one that checks that nothing is controlling the IDE (for `/control`) and one that checks that the current sender is the one that's controlling the IDE (e.g. `req.sender === PluginManager.controller.id`)

packages/selenium-ide/src/neo/stores/view/UiState.js
cdifino: Just want to make it more flexible to the user, if there are multiple instance within the same extension ID you want the user to be able to control what instance is talking to Selenium IDE.

packages/selenium-ide/src/api/v1/index.js
cdifino: I think that the idea is to block everything but /control. When Selenium is controlled by a specific extension, no other extension should be listening to events for security purposes. On the other hand, the only way to interrupt the connection is to request control through the /control call. If the user accepts to change the connection, the new extension will have control, and all the plugins will be unplugged.

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -24,81 +24,140 @@ import recordRouter from './record'
...
-    dependencies: req.dependencies,
-    jest: req.jest,
-    exports: req.exports,
-  }
-  Manager.registerPlugin(plugin)
-  res(true)
-})
+router.post(
+  '/register',
+  controlledOnly((req, res) => {
```
corevo: I think we should only block other extensions from calling `/control` right?

If the extension is in controlled mode, other plugins should still work, what do you think?

packages/selenium-ide/src/neo/stores/view/UiState.js
```diff
@@ -531,6 +533,16 @@ class UiState {
...
     this._project.saved()
   }
+
+  @action.bound
+  startConnection() {
+    this.isControlled = true
+  }
+
+  @action.bound
+  endConnection() {
```
corevo: No reason to add this, to break the control state the user will have to restart, so this will be wiped regardless.

packages/selenium-ide/src/neo/IO/filesystem.js
```diff
@@ -72,7 +72,16 @@ export function loadAsText(blob) {
 
 export function saveProject(_project) {
   const project = _project.toJS()
-  downloadProject(project)
+  if (UiState.isControlled) {
+    const saveMessage = {
+      action: 'event',
+      event: 'projectLoaded',
+      project: project,
+    }
```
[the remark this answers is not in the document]
cdifino: yes we can change the event type.

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -24,80 +25,148 @@ import recordRouter from './record'
...
+  controlledOnly((_req, res) => {
+    res({ id: UiState.project.id, name: UiState.project.name })
+  })
+)
+
+router.post('/control', (req, res) => {
+  if (
+    (Manager.controller &&
+      Manager.controller.connectionId != req.connectionId) ||
+    !Manager.controller
```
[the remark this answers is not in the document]
cdifino: when manager.controller is undefined it only means that there is no assisted control *yet*. This scenario is directed to the case when there is a connected mode but the connection id is a different one or when there is not connected mode and there is a potential for connecting.

packages/selenium-ide/src/background/background.js
```diff
@@ -181,17 +181,83 @@ browser.runtime.onConnect.addListener(function(m) {
   port = m
 })
 
+browser.runtime.onMessage.addListener(restart)
+
+function restart(message) {
+  if (
+    message.control &&
+    message.controller.connectionId &&
+    message.controller.id
```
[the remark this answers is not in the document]
cdifino: the senders. Thant is how the plugin manifesto works. The sender (or extension Id) turn out to be the plugin id.

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import browser from 'webextension-polyfill'
```
melode11: Seems like not necessary import?

melode11:
All the changes are done in our side. current state of the connected Mode PR is thoroughly tested by our testing team, including 10+ different corner cases are covered and all work as expected.

packages/selenium-ide/src/background/background.js
```diff
@@ -181,17 +181,87 @@ browser.runtime.onConnect.addListener(function(m) {
...
   (message, sender, sendResponse) => {
     if (!message.payload) {
       message.payload = {}
     }
-    message.payload.sender = sender.id
+
+    let payload = message.payload
+
+    payload.sender = sender.id
+    if (message.uri === '/_connect' || message.uri === '/_close') {
```
corevo: We could have an internal API that starts with `/private` instead of prefixing with `_`.

packages/selenium-ide/src/neo/components/PauseBanner/index.jsx
```diff
@@ -19,18 +19,28 @@ import React from 'react'
...
         </div>
-      </div>
-    ) : (
-      <div />
-    )
+      )
+    } else if (UiState.isControlled) {
+      return (
+        <div className="connected-mode">
+          <div className="connected-mode">
```
corevo: Why both `div`s have the same class attached?
Also `controlled-banner` instead of `connected-mode`.

packages/selenium-ide/src/plugin/manager.js
```diff
@@ -137,6 +138,10 @@ class PluginManager {
     return this.plugins.find(p => p.id === pluginId)
   }
 
+  unregisterAllPlugins() {
```
corevo: Unnecessary

docs/api/plugins/system.md
```diff
@@ -62,4 +62,23 @@ Loads a project into the IDE, as if the user opened it, if the user has unsaved
   project: JSON parsed side file
 }
 ```
+### `POST /control`
 
+Start a connection with a specific connection Id from another chrome extension. When this connection is accepted by the user, Selenium IDE restarts and registers the caller, and the extension takes exclusive control of Selenium IDE until user closes Selenium IDE or another connection is accepted. When this mode is on, the save to computer functionality gets overritten by sending the side file to the extension controlling Selenium IDE. When in this mode, the user needs to add connectionId variable to make other API calls such as `POST /project`, `GET /project`, `POST /log`, `POST /register`.
```
corevo: Why use `connectionId` instead of using the controller extension ID? seems less secure since this is a user-generated token rather than a browser secure token.

packages/selenium-ide/src/api/v1/index.js
```diff
@@ -24,79 +25,211 @@ import recordRouter from './record'
...
+    errorCode: 'MissingConnectionId',
+    error: 'No connection Id specified',
+  },
+  missingProject: {
+    errorCode: 'MissingProject',
+    error: 'Project is missing',
+  },
+}
+
+function checkControl(req) {
```
corevo: This function name is very ambiguous.

The function returns true in 2 cases
 - Extension is not controlled by anyone
 - Extension is controlled, and the request is sent by the controller

While returning false in 1 case
 - Extension is controlled by someone else

So I think we should split this function (and the decorator) into two that would be more readable
`isControlled` - would return true/false if the IDE is under control of any extension
`isControlledByThisExtension` - would return true only if the current extension is the controller

Then we can compose the decorators.

packages/selenium-ide/src/neo/IO/filesystem.js
```diff
@@ -82,15 +82,17 @@ function downloadProject(project) {
...
-      filename: sanitizeProjectName(project.name) + '.side',
-      url: createBlob(
-        'application/json',
-        beautify(JSON.stringify(project), { indent_size: 2 })
-      ),
-      saveAs: true,
-      conflictAction: 'overwrite',
-    })
+    return UiState.isControlled
+      ? Promise.resolve()
```
corevo: This should not silently fail, we should send a message to the controller that the user requested to save.

Look at [PlaybackState](https://github.com/SeleniumHQ/selenium-ide/blob/v3/packages/selenium-ide/src/neo/stores/view/PlaybackState.js#L291) to get an idea of how to send an event over to an extension.

I was thinking of something like:
```js
{
      action: 'event',
      event: 'saveProject',
      options: {
        project
      }
}
```

packages/selenium-ide/src/background/background.js
nbansal1993: This sounds good. I will do these changes and push them to PR

docs/api/plugins/system.md
melode11: Agree, we can switch to use sender id.

packages/selenium-ide/src/api/v1/index.js
melode11: I think I can make the name to be more clear. Split to more decorators sounds OK but we are never going to used the sparated decorators in other places. So rather keep it in a single function to simply the call hierarchy. Is that make sense?

melode11: Actually it is hard to think a good name for it. I'll just do split.

packages/selenium-ide/src/neo/IO/filesystem.js
melode11: We already send the plugin with the emit of the project entity. Do we expect to send another event here?

packages/selenium-ide/src/neo/components/PauseBanner/index.jsx
melode11: Yes, there's also a lot of duplications in these 2 css class. I will try to refactor the css class a little.

packages/selenium-ide/src/plugin/manager.js
melode11: > Unnecessary

Fixed.

packages/selenium-ide/src/api/v1/index.js
corevo: Its used on [line 88](https://github.com/SeleniumHQ/selenium-ide/pull/823/files#diff-1c12adce57bbfc73c193da228a08bd19R88)

packages/selenium-ide/src/background/background.js
```diff
@@ -181,17 +181,82 @@ browser.runtime.onConnect.addListener(function(m) {
...
+        verb: 'post',
+        payload: null,
+      })
+      .then(() => {
+        openPanel({ windowId: 0 }).then(() => {
+          var payload = { ...message }
+          delete payload.restart
+
+          const newMessage = {
+            uri: '/privateConnect',
```
corevo: I meant `/private/connect` and use the `/private/` prefix in general for internal IDE messages.

Also keep things lower cased.

packages/selenium-ide/src/neo/IO/filesystem.js
corevo: Emit project is something else, in controlled mode we shouldn't emit to begin with.

melode11: So to clarify. 
You want in control mode, IDE will send a saveProject event, do you still want to keep the emit? I guess we should still keep it for consistency between control and non-control mode?

Issue #844, opened 2019-10-07
C# code export (xUnit)


georgemjohnson11:
## 🚀 Feature Proposal
Reference: #592 

Code Export using C# and xUnit

## Motivation

I Work in a .NET environment and we needed to export to xUnit

## Example

This would be a similar package to the java-junit or python-pytest exports

PR #845, opened 2019-10-07
Code export C# xUnit

georgemjohnson11:

 Issue: #844


corevo:
Thanks for helping out with .net 🐢 

It might take me a bit to get to it, but I'll make sure to have a look.

packages/code-export-csharp-xunit/src/hook.js
```diff
@@ -0,0 +1,137 @@
...
+    result[hookName] = generate(hookName)
+  })
+  return result
+}
+
+function beforeEach() {
+  const params = {
+    startingSyntax: {
+      commands: [
+        { level: 0, statement: 'public class DriverFixture : IDisposable' },
```
georgemjohnson11: In proper usage, this should be a beforeAll() method to properly represent what this part of the code is doing up to line up to line 82. But I was trying to follow guidelines here which prefers beforeEach() so combined the two: https://www.seleniumhq.org/selenium-ide/docs/en/plugins/emitting-setup-teardown/
This also doesn't change functionality but for further reading on how xUnit passes along a shared context (the driver) using ClassFixtures: https://xunit.net/docs/shared-context

georgemjohnson11:
> Thanks for helping out with .net 🐢
> 
> It might take me a bit to get to it, but I'll make sure to have a look.

@corevo Thanks for the code review! I wrote comments on two nice-to-haves that I wasn't sure how to get past, but doesn't detriment too heavily to get most of this export ready for use. I'm in no rush at the moment since this has been working out to export my own test suites 😆 

The only things I haven't done are get the release ready as outlined here: https://github.com/SeleniumHQ/selenium-ide/wiki/Release-Process

packages/code-export-csharp-xunit/src/index.js
```diff
@@ -0,0 +1,122 @@
...
+// Create generators for dynamic string creation of primary entities (e.g., filename, methods, test, and suite)
+function generateTestDeclaration(name) {
+  return `[Fact]\npublic void ${exporter.parsers.capitalize(
+    exporter.parsers.sanitizeName(name)
+  )}() {`
+}
+function generateMethodDeclaration(name) {
+  return `public void ${exporter.parsers.sanitizeName(name)}() {`
+}
+
```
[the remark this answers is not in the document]
tourdedave: Name spacing isn't something we initially considered for code export. The goals were 1) export a target test or suite 2) with all commands working (e.g., recorded and manually added commands, stored variables, etc.), 3) with plugin support, 4) that is well structure, and 5) idiomatic to the target language. After that, if the user wanted more, they could make the changes themselves.

packages/code-export-csharp-xunit/src/hook.js
tourdedave: Good to know, thanks! Appreciate you being consistent with the other languages, though.

packages/code-export-csharp-xunit/src/hook.js
```diff
@@ -0,0 +1,127 @@
...
+        { level: 2, statement: 'driver.Dispose();' },
+        { level: 1, statement: '}' },
+        { level: 0, statement: '}' },
+      ],
+    },
+    endingSyntax: {
+      commands: [
+        {
+          level: 0,
+          statement: `public class TestSuite : IClassFixture<DriverFixture>`,
```
tourdedave: The suite name needs to be dynamic, which is why in the other languages it's handled in the index.js (e.g., `suiteDeclaration` passed to `emit.suite`). Also, this approach prevents the rest of the hooks from working, which mean it will break plugin support. Is it possible to refactor this into a structure that's similar to the Java code export?

tourdedave: Alternatively, you can look at the csharp-nunit branch I've been working on (e.g., [hooks](https://github.com/SeleniumHQ/selenium-ide/blob/code-export-csharp/packages/code-export-csharp-nunit/src/hook.js) and [index](https://github.com/SeleniumHQ/selenium-ide/blob/code-export-csharp/packages/code-export-csharp-nunit/src/index.js)).

We'll want to support both test frameworks, so there will likely be a common package for C# which will abstract the command emitters out.

georgemjohnson11: I've declared the suiteDeclaration again and simplified the implementation of xUnit. This now uses the beforeEach, afterEach, and declareVariables hooks, but xUnit doesn't have the concept of beforeAll and afterAll so I've omitted those. Unfortunately, the suite name is still static as xUnit needs the first method to be the name of the class for reusing a context (the driver). Would this work or does the suite name need to be unique and passed in a name?

tourdedave: What about [class fixtures](https://xunit.net/docs/shared-context#class-fixture)? Looks like they support running code once at the beginning and at the end of a suite. But it seems to be an either/or proposition. Is that right?

I hear you about the name being the same, but the hooks should not be responsible for declaring the suite. That's the purpose of `suiteDeclaration` in index. Hooks are meant to fill in the gaps within the suite and tests. We can forego the name being dynamic and just stick with it being hard-coded.

georgemjohnson11: Class fixtures are no longer relevant unfortunately. Utilitizing IDisposable, the class fixture supported both running code at the beginning and the end. However once I implemented this as the suite name, it didn't make sense to use class fixtures. The way ordered suites are emitted, the hooks names wouldn't match their functionality. So now I have a clean driver context for every test using a simple constructor and dispose.

To further elaborate on how this was implemented here using class fixtures is a recommended approach as just before the first tests in the TestSuits were run, xUnit.net would create an instance of DriverFixture. Then after all tests were ran, Dispose would be called due to the implementation of IDisposable. For each test, it would create a new instance of TestSuite(), and pass the shared instance of DriverFixture to the constructor.

tourdedave:
Just as any FYI, @georgemjohnson11  - I see why you wanted to introduce namespacing (e.g., ambiguity warnings). It didn't make the cut for this pass, but I think there's a simple way to do it if not offering it ends up being problematic for users. Since the beforeAll and afterAll hooks couldn't be used I pulled in the emitting template to index.js ([link](https://github.com/SeleniumHQ/selenium-ide/blob/v3/packages/code-export-csharp-xunit/src/index.js#L114-L144)) and removed those hooks from rendering. We could easily add dynamic namespace rendering to the template.

tourdedave:
I circled back and added namespacing to the export. Everything's getting released today in `v3.15.0`.

Thanks again!

PR #868, opened 2019-11-09
867 Add all commands to context menu

### LINKED ISSUE #867: Add all remaining commands to context menu
## 🚀 Feature Proposal

Add all available Assert, Store, Verify and Wait For commands to the context menu (the menu that is shown when you right click on a page)

## Motivation

Currently only a small subset of these commands is available in this menu. It would be good to be able to choose the full set so that you can easily add any of these commands while recording a test

## Example

For example, you cannot currently add an assertion to confirm that an element is present. You need to resort to workarounds like adding an assertion on the element text and then manually modifying it to an assertion on element present

carlos-granados:

Fixes #867 

Adds all available Assert, Store, Verify and Wait For commands to the context menu (the menu that is shown when you right click on a page)

Currently only a small subset of these commands is available in this menu. This code allows you to choose the full set so that you can easily add any of these commands while recording a test

Since adding all these commands to a single menu would create an unmanageable (too long) menu, we have modified the menu to have two levels. The first level defines the action to take (Assert, Store, Verify, Wait For) and the second level specifies the details. See this image:

Screen Shot 2019-11-09 at 07 20 06

We have added some comments to the PR to further explain what has been done

### DIFF
--- a/packages/selenium-ide/src/content/record.js
+++ b/packages/selenium-ide/src/content/record.js
@@ -644,10 +644,18 @@ Recorder.addEventHandler(
         let tmpText = bot.dom.getVisibleText(event.target)
         record(m.cmd, tmpTarget, tmpText)
       } else if (m.cmd.includes('Title')) {
-        let tmpTitle = goog.string.normalizeSpaces(event.target.ownerDocument.title)
+        let tmpTitle = goog.string.normalizeSpaces(
+          event.target.ownerDocument.title
+        )
         record(m.cmd, [[tmpTitle]], '')
-      } else if (m.cmd.includes('Present') || m.cmd.includes('Checked') || m.cmd.includes('Editable')
-          || m.cmd.includes('Selected') || m.cmd.includes('Visible') || m.cmd === 'mouseOver') {
+      } else if (
+        m.cmd.includes('Present') ||
+        m.cmd.includes('Checked') ||
+        m.cmd.includes('Editable') ||
+        m.cmd.includes('Selected') ||
+        m.cmd.includes('Visible') ||
+        m.cmd === 'mouseOver'
+      ) {
         record(m.cmd, tmpTarget, '')
       } else if (m.cmd.includes('Value')) {
         let tmpValue = event.target.value
--- a/packages/selenium-ide/src/content/record.js
+++ b/packages/selenium-ide/src/content/record.js
@@ -638,18 +638,20 @@ Recorder.addEventHandler(
   'contextmenu',
   function(event) {
     let myPort = browser.runtime.connect()
-    let tmpText = locatorBuilders.buildAll(event.target)
-    let tmpVal = bot.dom.getVisibleText(event.target)
-    let tmpTitle = goog.string.normalizeSpaces(event.target.ownerDocument.title)
+    let tmpTarget = locatorBuilders.buildAll(event.target)
     myPort.onMessage.addListener(function(m) {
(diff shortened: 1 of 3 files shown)


packages/selenium-ide/src/content/record.js
```diff
@@ -638,18 +638,20 @@ Recorder.addEventHandler(
   'contextmenu',
   function(event) {
     let myPort = browser.runtime.connect()
-    let tmpText = locatorBuilders.buildAll(event.target)
```
carlos-granados: We are actually calculating the target not getting the text of the element, so the name of this variable was misleading, that's why I changed it

packages/selenium-ide/src/content/record.js
```diff
@@ -638,18 +638,20 @@ Recorder.addEventHandler(
...
-    let tmpVal = bot.dom.getVisibleText(event.target)
-    let tmpTitle = goog.string.normalizeSpaces(event.target.ownerDocument.title)
```
carlos-granados: These two variables are only needed for some specific commands so I have moved them to the corresponding cases

packages/selenium-ide/src/neo/IO/SideeX/recorder.js
```diff
@@ -381,12 +381,12 @@ export default class BackgroundRecorder {
       typeof message.value === 'undefined'
     ) {
       logger.error(
-        "This element does not have property 'value'. Please change to use storeText command."
+        "This element does not have property 'Value'. Please change to use a 'Text' command instead."
```
carlos-granados: These messages mentioned the store command but will be also shown for assert and verify commands, so I have changed them to be more generic

packages/selenium-ide/src/neo/side-effects/contextMenu.js
```diff
@@ -22,40 +22,273 @@ function createContextMenus() {
     contexts: ['all'],
   })
   browser.contextMenus.create({
-    id: 'verifyText',
-    title: 'Verify Text',
+    id: 'assert',
```
carlos-granados: This is the parent menu item

packages/selenium-ide/src/neo/side-effects/contextMenu.js
```diff
@@ -22,40 +22,273 @@ function createContextMenus() {
...
+    title: 'Assert',
     documentUrlPatterns: ['<all_urls>'],
     contexts: ['all'],
   })
   browser.contextMenus.create({
-    id: 'verifyTitle',
-    title: 'Verify Title',
+    id: 'assertChecked',
```
carlos-granados: And these are all the child menu items

packages/selenium-ide/src/neo/side-effects/contextMenu.js
```diff
@@ -22,40 +22,273 @@ function createContextMenus() {
...
+  })
+   browser.contextMenus.create({
+    id: 'verifyValue',
+    title: 'Value',
+    documentUrlPatterns: ['<all_urls>'],
+    contexts: ['all'],
+    parentId: 'verify'
+  })
+  browser.contextMenus.create({
+    id: 'waitFor',
```
carlos-granados: Also added all the WaitFor commands which were not previously available in the context meny

tourdedave:
👏 Thanks! This is a great PR. Sorry for the delay in reviewing.

PR #870, opened 2019-11-10
869 Add WaitForText command

### LINKED ISSUE #869: Add a WaitForText command
## 🚀 Feature Proposal

Add a new wait command that will wait until the text of an element has a specific value.

## Motivation

Sometimes you need to wait until the specific text of an element has a certain value, usually because this text is set by a javascript process, maybe an ajax call. The current functionality of being able for an element to be present or visible is not enough as the element may be present and visible but its text may not have changed yet

## Example

Suppose that you had an editor that lets you edit some properties of a document, for example its title. You do this by clicking on an "edit title" button that loads a javascript dialog where you enter the new title. When you click save, the title is sent to the backend using ajax and once the request finishes correctly, you update the title in the page. If you wanted to write a test to confirm that the title is changed in the page, if you try to assert that the title matches the new title after submitting the dialog it will fail as the title would still have the old title until the ajax request finishes. And since the title is already there you cannot just wait for it to be present or visible before making the assertion

carlos-granados:

Fixes #869 

Add a new `waitForText` command that will wait until the text of an element has a specific value

I have verified that the functionality works in the IDE and all the tests are passing. I also modified all the code exporters and, though the exported code looks ok, I have not been able to actually verify that it works fine, will probably need that whoever has worked in each particular exporter takes a look to confirm that the output is actually correct

Also, the other wait commands used the value parameter to pass the timeout of the wait. Since we now need to use this parameter to pass the text that needs to be checked, we cannot pass this timeout so I had to use the default timeout


carlos-granados:
There is one test that is failing in Travis but I can't understand why. It is the Ruby exporter, somehow the exporter is emitting the css locator surrounded by single quotes instead of double quotes. Locally all the tests pass and when I test the exporter it is using double quotes as it should. I have no idea why this specific test is emitting something different and failing

tourdedave:
That is strange. I updated the code and snapshots for that earlier today. I'll take a look.

carlos-granados:
It was due to this commit: https://github.com/SeleniumHQ/selenium-ide/commit/26a2788b1bbc5207b87c60987c6c80dc519386b7 where you updated the quotes used by the Ruby exporter. I have updated our fork with the new base code, updated the snapshot and now everything is passing

tourdedave:
💯

PR #926, opened 2020-01-08
925 properly save the handlers so that they are correctly removed

### LINKED ISSUE #925: Stopping and restarting the recorder adds event listener twice
## 🐛 Bug Report

When recording, if you stop recording and then start recording again, the event listeners are not properly removed, so we end up duplicating them, which results in recording commands more than once

## To Reproduce

For example, start recording a test and select an item on a select box. Stop recording and restart recording. select another item on the select box. The select command will be recorder twice and duplicated. If you stop the recorder and restart it once more and then select another item, then this will be recorded three times and so on

## Expected behavior

If we stop recording and start recording again, the commands should not be duplicated

## Project file reproducing this issue (highly encouraged)

I reproduced this by going to google, typing something in the search box, stopping the recorder, starting it again and typing something else in the search box. The file is named `side.txt` because Github did not let me add a file with the .side extension

[test.side.txt](https://github.com/SeleniumHQ/selenium-ide/files/4035912/test.side.txt)

## Environment

OS: OSX
Selenium IDE Version:  3.16.1
Selenium SIDE Runner Version:  n/a
Browser: Chrome
Browser Version: 79.0.3945.88

carlos-granados:
Properly save the event handlers so that they can be removed when recording is stopped
Fixes #925 

### Description
When setting the event handlers we were adding them with `.bind(this)` but then we saved them without this, this made the signature of the event handler different and when trying to remove them they would not be removed

### Motivation and Context
When recording, if you stop recording and then start recording again, the event listeners are not properly removed, so we end up duplicating them, which results in recording commands more than once

### Types of changes

### Checklist
- [X] All new and existing tests passed.

### DIFF
--- a/packages/selenium-ide/src/content/record-api.js
+++ b/packages/selenium-ide/src/content/record-api.js
@@ -102,12 +102,9 @@ Recorder.prototype.attach = function() {
       const handlers = Recorder.eventHandlers[eventKey]
       this.eventListeners[eventKey] = []
       for (let i = 0; i < handlers.length; i++) {
-        this.window.document.addEventListener(
-          eventName,
-          handlers[i].bind(this),
-          capture
-        )
-        this.eventListeners[eventKey].push(handlers[i])
+        let handler = handlers[i].bind(this)
+        this.window.document.addEventListener(eventName, handler, capture)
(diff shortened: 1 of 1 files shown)


tourdedave:
💯

Issue #965, opened 2020-02-24
Support Accessibility regarding Coloring and Design


ansarmie:
## 🐛 Bug Report

Fulfill the **W3 Standards** **1.4.3 - Contrast (Minimum)** and **1.4.11 - Non-text Contrast** (Recommendation) to support the use of the application by people with Hard of sight and Keyboard or Screen Reader users.

More info here:

- [Guideline 1.4 Distinguishable](https://www.w3.org/TR/WCAG21/#distinguishable)
- [Understanding Success Criterion 1.4.3: Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html)
- [Understanding Success Criterion 1.4.11: Non-text Contrast](https://www.w3.org/WAI/wcag21/understanding/non-text-contrast.html)

## To Reproduce

- Color contrast ratio in Name your project and Controlled by Selenium IDE is less than 4.5:1.

## Expected behavior

- The text in a control, label or title should be at least of 4.5:1.

## Project file reproducing this issue (highly encouraged)

This is a issue with the UI, not of a specific Selenium scenario.

## Environment

OS: Windows 10
Selenium IDE Version:  3.16.1
-->
Browser: Google Chrome
Browser Version: 80.0.3987.116 (Official Build) (64-bit)

tourdedave:
Fixed in [v3.17.0](https://github.com/SeleniumHQ/selenium-ide/releases/tag/v3.17.0)

PR #972, opened 2020-02-28
accessibility fixes - color contrast, focus indicator, and tab order

tonglingchen:
### Description
This is to fix issue #965 and #975. Changes include:
1. Use the default outline to indicate that control is on focus
2. Update banner background to have a better contrast against the background
3. Update color for several icons to have a better contrast against the background
4. Swap buttons on dialog to fix the tab order for Windows
5. Set a min-height for main container and show scrollbar when needed

### Motivation and Context
Fulfill the W3 Standards 1.4.3 - Contrast (Minimum) and 1.4.11 - Non-text Contrast (Recommendation) to support the use of the application by people with Hard of sight and Keyboard or Screen Reader users.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist


packages/selenium-ide/src/neo/components/ActionButtons/ActionButton/style.css
```diff
@@ -4,7 +4,6 @@
   height: 24px;
   padding: 0;
   margin: 1px 0;
-  outline: 0;
```
corevo: I like the outline that was added to the open button, instead of using the default one how about we keep `outline: 0`, and add to the `:focus` a `outline: auto 1px -webkit-focus-ring-color;`?

packages/selenium-ide/src/neo/components/Test/style.css
```diff
@@ -45,11 +45,8 @@
 }
 
 .test > a.selected {
-  background-color: var(--selected-color);
-}
-
-.test:focus > a.selected {
-  background-color: var(--active-color);
+  background-color: #5c5c5c;
```
corevo: This color makes the more button (visible on the right on hover), barely visible, either we can change the button's color to white as well, or find a different color.

packages/selenium-ide/src/neo/containers/Panel/index.jsx
```diff
@@ -55,10 +55,13 @@ if (!isTest) {
 
 if (userAgent.os.name === 'Windows') {
   require('../../styles/conditional/scrollbar.css')
-  require('../../styles/conditional/button-direction.css')
   require('../../styles/conditional/text.css')
 }
 
+if (userAgent.os.name === 'Mac OS') {
+  require('../../styles/conditional/button-direction.css')
```
corevo: What is the expected behavior on each OS?

On Mac the button order is `cancel | action` and on everything else `action | cancel`?

What specs the order on each OS?

packages/selenium-ide/src/neo/components/ActionButtons/ActionButton/style.css
tonglingchen: To be consistent across all controls? I can do that but still prefer the keep the default style. Using default means less thing to worry 😃

packages/selenium-ide/src/neo/components/Test/style.css
tonglingchen: sorry missed this one. will update the button's color.

packages/selenium-ide/src/neo/containers/Panel/index.jsx
tonglingchen: I am thinking it should be on Windows it is `action | cancel` and or o ther OS it is `cancel | action`. Don't think there is a strict spec to follow. 
Will update that `if (userAgent.os.name === 'Mac OS')` check.

packages/selenium-ide/src/neo/components/ActionButtons/ActionButton/style.css
tonglingchen: @corevo do you think we still need to reset outline?

corevo: Isn't the default `auto medium -webkit-focus-ring-color`?

But if we do get into it, we can specify something specific, we have the hover color `#40a6ff`, what do you think of `auto 1px #40a6ff`?

packages/selenium-ide/src/neo/containers/Panel/index.jsx
corevo: It did used to do that, and that CSS rule swapped the order of buttons on Windows, treating Windows as the exception rather than the rule.

Is this change strictly for tab order? if so the problem was pushed from Windows back to Mac AFAIK.

packages/selenium-ide/src/neo/components/ActionButtons/ActionButton/style.css
tonglingchen: Sure I will make this change.

packages/selenium-ide/src/neo/containers/Panel/index.jsx
tonglingchen: Yes you are right. I am thinking about removing that CSS button direction rule so buttons have same order across OSes. What do you think the button order should be - `action | cancel` or `cancel | action`? It would be great if we know the user base distribution - how many users are on Windows, how many users are on MacOS, etc. Thus we can pick an order which works for majority of users. But I doubt there is such telemtry 😃.

packages/selenium-ide/src/neo/components/ActionButtons/ActionButton/style.css
tonglingchen: @corevo cannot use `auto` as the style because it will just fallback to default style. Now the outline is `dashed 1px #4d90fe` when focused. I changed the color because `#40a6ff` fails contrast ratio check.

corevo: Really, on what browser? I tested on Chromium 80 and was able to keep `outline-style: auto` and set `outline-color` & `outline-width`, but if this is the case then `outline-style: solid` should be good.

corevo:
I was thinking about button directly, instead of changing the direction, what if instead we place a variable (could be global 🤔) that would just tell us the button order, `normal` or `reversed`.

For `normal` it would be `cancel | action`
For `reversed` it would be `action | cancel`

And based on the user agent we would render them in the correct order, instead of changing the style, that would seem in fit with the user, and the tab order will be in sync on all agents.

Thoughts?

packages/selenium-ide/src/neo/components/ActionButtons/ActionButton/style.css
tonglingchen: For me if `outline-style` is `auto`, the `outline-width` is ignored. It renders a outline thicker than 1px. 
Changed to `solid`.

tonglingchen:
> I was thinking about button direction, instead of changing the direction, what if instead we place a variable (could be global 🤔) that would just tell us the button order, `normal` or `reversed`.
> 
> For `normal` it would be `cancel | action`
> For `reversed` it would be `action | cancel`
> 
> And based on the user agent we would render them in the correct order, instead of changing the style, that would seem in fit with the user, and the tab order will be in sync on all agents.
> 
> Thoughts?
> 
> EDIT: my reasoning of global is that it is an extension of the agent, therefore a global, but we could inject it or use a store w/e.

tonglingchen:
@corevo I have pushed change to fix the button direction.

tourdedave:
@corevo thoughts?

packages/selenium-ide/src/neo/components/Dialogs/Dialog/index.jsx
```diff
@@ -54,11 +54,11 @@ export default class DialogContainer extends React.Component {
     if (!this.props.buttons) {
       return null
     }
-    var buttons = (UiState.dialogButtonDirection === 'reversed'
+    const buttons = (UiState.dialogButtonDirection === 'reversed'
       ? this.props.buttons.slice().reverse()
       : this.props.buttons
     ).filter(button => button !== null)
-    return <div className="right">{buttons}</div>
```
tonglingchen: This is needed because it shifts buttons to the right side of the dialog.

PR #1042, opened 2020-05-08
fix code-exports with unnecessary flags

toddtarsi:
I have some commands with the following syntax:

https://ibb.co/D9WDK5K

NOTE: This syntax is generated using the verifyElementPresent command from the side Chrome extension.

Because commands like verifyElementPresent don't have any processing for a value flag, it detonates the export pipeline. This resolves that. Thanks for the great repo!

### Motivation and Context

Make code export a bit more flexible with the side file inputs vs side-models.
If the commandSchema doesn't even specify a target / value, then I think we can assume its optional.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist


toddtarsi:
Hey all, I made sure all linting here passes now, sorry about that.

I just wanted to say that I think this is a very important PR in the short term, because code export of step files to mocha javascript via the side-utils package doesn't work for me when using very basic assertions like verifyElementPresent.

Also, I wanted to say thank you very much for the great packages and keeping work going on the selenium suite. Using Selenium with modern tooling is great. I worked with selenium ui back in the day and am glad to see the ecosystem still progressing, albeit it sucks how changing browser internals often force a major re-roll of this system every five years or so. Please let me know if I can help with anything in your tooling.

AutomatedTester (approved):

kruus:
I was unable to export anything today until I **commented the 'echo' commands** I'd used for debug.

Issue #1060, opened 2020-06-02
selenium-ide/packages/code-export-java-junit/src/command.js#emitRunScript '${generateScriptArguments(script)}' must be outside double quotes


etyero:
## 🐛 Bug Report

selenium-ide/packages/code-export-java-junit/src/command.js#emitRunScript 

'${generateScriptArguments(script)}' must be outside double quotes
![image](https://user-images.githubusercontent.com/26212853/83473117-68a98080-a4bb-11ea-938d-0a50fc395914.png)

tourdedave:
Thanks for the heads up. But I'm curious why you're sending these as one-off issues and not submitting a pull request to fix it?

etyero:
> Thanks for the heads up. But I'm curious why you're sending these as one-off issues and not submitting a pull request to fix it?

Sorry，I‘m a beginner，I will fix these issues next time

Issue #1059, opened 2020-06-01
selenium-ide/packages/code-export-java-junit/src/command.js#emitControlFlowForEach  Why minus 1 ??


etyero:
## 🐛 Bug Report

selenium-ide/packages/code-export-java-junit/src/command.js#emitControlFlowForEach 
Why minus 1 ?

![image](https://user-images.githubusercontent.com/26212853/83396050-eb83fa00-a42d-11ea-8398-5c14577e16a1.png)

tourdedave:
If I recall correctly it was to avoid an IndexOutOfBounds exception.

Please note that the issue tracker is not for questions. Please follow the issue template.

etyero:
> IndexOutOfBounds

But can't get the last element like this?

tourdedave:
Worth double-checking.

Issue #1058, opened 2020-06-01
selenium-ide/packages/code-export-java-junit/src/command.js#emitControlFlowForEach's iteratorVarName is wrong


etyero:
## 🐛 Bug Report
selenium-ide/packages/code-export-java-junit/src/command.js#emitControlFlowForEach's iteratorVarName is wrong:
![image](https://user-images.githubusercontent.com/26212853/83393891-26842e80-a42a-11ea-963d-23d57e7718db.png)

tourdedave:
Pull requests welcome.

PR #1061, opened 2020-06-02
Update command.js

etyero:
Fix issue #1060,#1059,#1058

### DIFF
--- a/packages/code-export-java-junit/src/command.js
+++ b/packages/code-export-java-junit/src/command.js
@@ -301,11 +301,11 @@ function emitControlFlowForEach(collectionVarName, iteratorVarName) {
       },
       {
         level: 0,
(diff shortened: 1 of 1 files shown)


toddtarsi:
@etyero - This stuff looks good to me, but can you pull in the latest from v3? That will make it so that our new test system is able to run and ensure all snapshots look okay, or will need to be updated as needed or whatever. Better safe than sorry after all.

toddtarsi (approved):
Everything here LGTM 👍 

I'm probably not a highly competent reviewer on this. I haven't used the java bindings before, but all and all it looks correct.

That script thing should probably have the script as a string, and then some other things that make the comma delimited list so that things aren't wacky.

Also, the variable name should be the same there as it is at the other spots probably. 

Nice work 👍

PR #1098, opened 2020-07-24
[Build] Initial Bazel setup

AutomatedTester:
This is a draft PR and will like land directly but sharing if people want to build on top


AutomatedTester:
landed directly so closing

Issue #1145, opened 2020-10-15
Export to Django Live Server Format


mkalioby:
## 🚀 Feature Proposal

Django (Web Framework for Python) has a selenium driver that avoids starting the HTTP Server for selenium to run, Django start the server on a random port and the address is available in the test class.

Refer to [Django Live Server](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#liveservertestcase)

## Motivation

The format is so close to PyTest, only minor modifications are required and the developers can create their test cases by Selenium IDE and export to Django Server format then run the tests by 

`python manage.py test` 

without starting the http server, which is much easier for CI pipelines.

corevo:
We strictly only provide support for first party (or most common) test framework in each supported language, that is to limit both maintenance of the project, and to prevent creating a precedent where we would write an export for every setup available.

Because of that we've created plugins, I highly suggest writing this as a plugin for Selenium IDE, as we will not accept this PR.

mkalioby:
Thanks for your reply.

Can you direct me for the plugin documentation?

Thanks in advance.

corevo:
Of course, its available with the rest of the docs on the [official webpage](https://www.selenium.dev/selenium-ide/docs/en/plugins/code-export).

PR #1146, opened 2020-10-15
Export in Django Live Server format

mkalioby:
### Description
Created new export plugin for Python Django Live Server See #1145 

### Motivation and Context
Export to a new format which is close to pytest with minor modifications

### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Checklist


toddtarsi:
@mkalioby - This looks good to me, but I have a couple requests if alright:

1. Can you drop the yarn.lock file? We don't tend to keep those in source in the sub-packages. Even if it is good form, it's more of an OCD thing.
2. Can you add test snapshots similar to the javascript_mocha library? These make it so that long term, we'll have to worry about jacking things up on django live server when making changes to the test-exporter.
3. Can you pull in the latest code here and deal with the conflicts?

Thanks for the hard work, and sorry for the delay!

mkalioby:
Sure, I can do that but the problem that truck now in the electron version not the browser plugin version, Can you please where is the code export module now?

Thanks

toddtarsi:
@mkalioby - After talking with the team a bit, it sounds like we're not able to add this option because Python / Django Live Server is not an officially supported binding. As a result, we won't be able to merge this into the core repo and put the onus of long term support on us. I'm sorry for the mistakes, I'm rather new here! However, you could either make it a standalone repo to help others in a similar place, or a plugin to make it so others could use your code as an extension to add Python / Django Live Server code export:

https://www.selenium.dev/selenium-ide/docs/en/plugins/code-export

Please let me know if I can help with either one of these. Thanks for helping add to the ecosystem, and sorry again about the confusion!

PR #1223, opened 2021-04-07
Render method declarations similarly to tests

toddtarsi:
### Description

This makes method body generation work more similarly to test generation.

Rather than using a one-off method to generate our method declarations, we take in the render and testLevel variables from emitTests, and use these to generate our method declarations.  

### Motivation and Context

The way it previously worked, command emitters that used any syntax other than raw strings would render as [Object object] within method declarations.
Standard emitters like this wouldn't work:

```
click: async function emitClick() {
  return promise.resolve({ commands: [{ statement: 'xyz.click()' level: 0 }] })
}
```

This meant that re-using tests as methods required some pretty heavy changes to the previously written export format.
Additionally, this meant that commands that affected indentation in methods, like control flow blocks, would generally fail to render correctly due to their usage of staringLevelAdjustment and endingLevelAdjustment.

### Types of changes

### Checklist
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/packages/code-export-python-pytest/__test__/src/__snapshots__/index.spec.js.snap
+++ b/packages/code-export-python-pytest/__test__/src/__snapshots__/index.spec.js.snap
@@ -140,7 +140,8 @@ class TestLogin():
     self.driver.find_element(By.LINK_TEXT, \\"Form Authentication\\").click()
     self.driver.find_element(By.ID, \\"username\\").send_keys(self.vars[\\"username\\"])
     self.driver.find_element(By.ID, \\"password\\").send_keys(self.vars[\\"password\\"])
-    self.driver.find_element(By.CSS_SELECTOR, \\".fa\\").click()
+    if self.driver.execute_script(\\"return (true)\\"):
+      self.driver.find_element(By.CSS_SELECTOR, \\".fa\\").click()
   
   def test_invalidcredentials(self):
     self.vars[\\"username\\"] = \\"blah\\"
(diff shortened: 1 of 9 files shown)


toddtarsi:
Also, I'm not sure what's going on with the CLA stuff. I signed it so long ago that it's saying I haven't signed it or something? When I go to sign it again, it tells me I already signed.

toddtarsi:
> Also, I'm not sure what's going on with the CLA stuff. I signed it so long ago that it's saying I haven't signed it or something? When I go to sign it again, it tells me I already signed.

NM, I needed to set my email in my local git config. CLA stuff passes now, thanks!

toddtarsi:
Looking at this option further, it seems like I probably want something for method indentation.

Also, if this is preferred, I can point this PR at master instead since it deals more with side-utils than the ide.

packages/code-export-python-pytest/__test__/src/__snapshots__/index.spec.js.snap
```diff
@@ -305,7 +306,8 @@ class TestValidcredentials():
     self.driver.find_element(By.LINK_TEXT, \\"Form Authentication\\").click()
     self.driver.find_element(By.ID, \\"username\\").send_keys(self.vars[\\"username\\"])
     self.driver.find_element(By.ID, \\"password\\").send_keys(self.vars[\\"password\\"])
-    self.driver.find_element(By.CSS_SELECTOR, \\".fa\\").click()
+    if self.driver.execute_script(\\"return (true)\\"):
```
AutomatedTester, line 309: Could you explain why we're doing this now? I don't understand this change

toddtarsi: @AutomatedTester - Yessir, the reason I added this logic is that my fixes target control flow logic statements more than anything else.

If you add control flow logic to a reused method currently, the if and end blocks will render as `[object Object]`, breaking code export in all languages. You can try it by using the updated side file in this branch in master, and export to any language. It will be 100% broken.

Not to get too long winded, but this is because method bodies were generated using a one-off method that didn't handle all cases like emitCommands does.

So I figured, in addition to writing these fixes, I would expand the test SIDE files so that control flow logic were included, and any future regressions would be caught. This was the simplest possible implementation of control flow logic:

```
if true
    click xyz
end
```

After adjusting the side file, the snapshots were failing because the new steps weren't there, so I had to update the snapshots as well.

Okay, that was a long explanation, but yeah.

AutomatedTester (approved):

PR #1239, opened 2021-05-21
WIP: Selenium IDE V4 via wrapping Selenium IDE V3

toddtarsi:
This is a general bucket for some work I've been doing to take Selenium-IDE v4 forward.

### Description

There are a number of things here, split into various commits. The commits can be cherry picked out individually and incorporated into master. The goal of this PR description is to take a large, vague PR and explain each commit directly.

* Swaps out the manual Chromedriver install process to use electron-chromedriver, which automatically syncs electron and chromedriver versions and allows us to easily come up a number of major versions in both electron and chromedriver. Commits: a180791 e27a3cf

* Adds a wrapper around Selenium IDE v3 as a package. My goal here is to use the loadExtension API and render the IDE on a tab. However, it might be better to do this a different way or give up on it. I'm still exploring this. The first time this is installed, it makes the build nightmarishly long (due to an early build step of running peru sync and then yarn in the github install), and then it gets faster after that. Commits: fccfb15 1aa6cb5

* Add electron-tabs to the frontend. This provides us a tab solution without having to roll it ourselves. This is currently using a github fork, while I wait for this [PR](https://github.com/brrd/electron-tabs/pull/120) to be merged. Once that is merged and the release is incremented, I'll switch to that. In addition, this moves the renderer script into a separate package, so the concerns of the primary webview aren't comingled with the concerns of the electron base. Commits: 747ab5e

* Set `nodeIntegration: false` and look at handling elevated operations from the main webview using an IPC channel. Commits: e8fe6b6

### Motivation and Context

* Swapping out the Chromedriver install method allows us to easily upgrade electron versions, and removes a challenging portion of the build process from being our concern.

* Using Selenium V3 within Electron, if at all feasible, will allow us to progress rapidly on v4, and also get double value out of any concurrent improvements or fixes in v3. There's a lot of good code, and I'd rather restart as little as possible.

* We needed a tab solution, and electron-tabs was pre-existing and already made for this. My only concern was that I thought that the `nodeIntegration` flag was required to be true. Also, moving the renderer out to a separate package frees us up to approach it more directly with front end technologies like webpack (and do things in typescript to maintain full hipness).

* Using `nodeIntegration: false` is important when working with outside sources. Any google search can pull up that information. As we're pretty much wrapping a pseudo-browser, outside sources are going to be something we're nearly always interacting with. Generally, doing whatever we can from this [template](https://github.com/reZach/secure-electron-template ) will be important, especially since much of what we're doing by design goes against many of these principals.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist


toddtarsi:
Closing this, as writing v4 as a new thing has been a lot less painful than trying to polyfill the extension api

PR #1295, opened 2021-11-12
Fix exports with multiple reused tests

### LINKED ISSUE #1291: Exports with run commands only exports first sub methods
## 🐛 Bug Report

When a Test Case has more than one use of the "run" command, the exportation of the code only includes the first sub tests, ommiting the ones from the second and beyond sub tests.

## To Reproduce

Create two Test Cases (TC01.1 and TC01.2) with any command on them ("echo" command for example). Then create another test case (TC01) and call them by using the "run" command. They should be as follow:   
run   |   TC01.1
run   |   TC01.2

Execute the last Test Case (TC01). Observe that in the Selenium IDE they run correctly showing the echo texts in the log window.

No export the last Test Case (TC01) to any languaje (Python for example). Mark the first two options and then save the exported file.

Edit the exported file and look the content.

## Expected behavior

The code should contain the principal module of the main Test Case:   
    *def` test_tC01(self):    
    # Test name: TC01   
    # Step # | name | target | value
    # 1 | run | TC01.1 |    
    self.TC011()   
    # 2 | run | TC01.2 |    
    self.TC012()*   
 

And also the definitions of each sub test Case

 *def tC011(self):
    print("Running Test Case 01. Sub Test 01")*

and 
   *def tC012(self):
    print("Running Test Case 01. Sub Test 02")*

## Observed Behaviour

The second method and any other method beyond the first "run" command are missing. The following lines are NOT included in the exportation 
   *def tC012(self):
    print("Running Test Case 01. Sub Test 02")*

## Project file reproducing this issue (highly encouraged)

Side file with an example of the issue   
[Run Command Bug Report.zip](https://github.com/SeleniumHQ/selenium-ide/files/7489190/Run.Command.Bug.Report.zip)

Python file with the exportation showing the issue
[test_tC01.zip](https://github.com/SeleniumHQ/selenium-ide/files/7489192/test_tC01.zip)

## Environment

OS: Ubuntu 18.04 
Selenium IDE Version:  v3.17.0 
Selenium SIDE Runner Version:  Unknown. Latest? 
Node version: 
Browser: Chrome
Browser Version: 95.0.4638.69 (Official Build) (64-bit)

toddtarsi:
Resolves #1291


diemol (approved):
Thank you, @toddtarsi!

Issue #1240, opened 2021-05-25
Can not use a variable containing a single quote


redescaflown:
## 🐛 Bug Report

I try to return a variable containing a single quote but I have a bug. Its because selenium-side-runner doesnt escape my single quote : I have this error : javascript error: Unexpected identifier'

If I realise the same test with a text without a single quote, all is Ok.

![escape](https://user-images.githubusercontent.com/17674866/119486898-0810f480-bd59-11eb-9fcb-5c5a2b4033bf.PNG)

Is there any way to force selenium-side-runner escaping this carachter ? 

I would like to precise that I am using selenium-side-runner command, not selenium IDE, it works fine in selenium IDE

kogbill:
I have a basic approach to solve this here:
https://github.com/kogbill/selenium-ide/commit/b5065585fdf72ab4c4f9c2f43f5eab67b5607007

toddtarsi:
This should have been resolved for a while. Tested in v4:
<img width="699" alt="Screen Shot 2022-11-13 at 7 31 59 PM" src="https://user-images.githubusercontent.com/4707505/201557224-9f302861-5631-4db7-ad67-bb502fdf0753.png">

PR #1297, opened 2021-11-13
#1240 facility for escaping variables in emitted code

toddtarsi:
This is just directly transporting the hard work of @kogbill and/or @zot into a version that is more merge-friendly.

Thanks again for your hard work!


toddtarsi:
@zot - I like the work here, but there are a couple things I'd like to think about before I merge this:

1. Is this where this should live? I'd like to make sure code-export is able to handle quote nesting as well, and to look at these two systems and see if one of these drives the other.

2. Does this handle all of the use cases? Rather than hardcoding for a specific amount of slashes we'd expect to encounter, I'd expect a capturing group regex that would effectively "increment". I want to make sure that, if we're here right now, we are escaping all the quotes, even the escaped quotes within the escaped quotes within the escaped quotes.

3. This is code I'd like to have a test for to ensure it works correctly, because it is too complex to tell immediately from here if this fully matches what we're expecting.

If you'd like to field these, I'd be supportive of working further with you. If not, I could take a look this weekend. That sound cool?

toddtarsi:
@zot - Also, I'm not sure anything here is wrong btw. I just want to look around and make sure this isn't missing functionality on the other half of our runner systems. I have to use some creative regexes in my personal use cases, so maybe something there could be tuned up a bit as well.

toddtarsi:
@zot - I was looking a bit closer here, and I don't think this is exactly where we want this fix.

So, the selenium-side-runner and selenize kind of sit on top of a particular flavor of code-export, and it is supposed to do exactly what you want here:

https://github.com/SeleniumHQ/selenium-ide/blob/v3/packages/side-utils/src/string-escape/index.js

This is what I was looking for. This handles nesting all the way down by incrementing the escape level and such.
If it doesn't, we should take a look here and resolve it to. Perhaps this needs to effectively remove newlines, run the regex, and them add them back.

I don't want to solve this with an escape hatch because I'd like to ensure this is also working for all the Python, Java, and other supported flavors.

zot:
This was just something I did at work because tests that worked in the browser didn't work in circleci. I wanted to contribute it to you guys so other people could benefit. Please feel free to use it as you see fit -- or not, if there are better options!

toddtarsi:
@zot - Is there any chance I could ask you for a screencap of the command that fails?
I have this, but I'm not sure it covers the requirements of what you're seeing:

Screen Shot 2021-11-16 at 10 31 50 AM

If this is going to be sufficient to see these issues on the selenium-side-runner output, cool and I'll add it to our stock tests for all language outputs. If not, please provide a snapshot of the step you're seeing this on and I'll use it as the basis for a test of all exports.

zot:
The failing command is "edit content". The code generator does not escape the value.

image

Here's a test that runs in the browser and fails in the runner. 

[quotey.zip](https://github.com/SeleniumHQ/selenium-ide/files/7561589/quotey.zip)

CSS and xpath locators fail in both the IDE and the runner when you use variables containing quotes. For example, `css=[content*="${quoteyThing}"]`. CSS can escape characters but I think xpath is more of a problem and might not be fixable without injecting a library into browsers.

toddtarsi:
@zot - You are a champion. I'll make sure something like this becomes a core test on all supported export languages. Thank you for finding this, explaining it, creating a fix, and everything. This is my big OS priority this weekend. 👍

zot:
No problem! I'm not positive I found all the places in the code that expand variables, though.

Also, those css/xpath substitutions are still a problem both in the browser and the runner. I hope you guys can fix those issues -- sorry I haven't made an actual Git issue about that part...

toddtarsi:
@zot - Good call out with the selectors. Xpath will unfortunately be a big problem because it bleeds into the string quoting system itself, which gets into certain language level things. For something like that, we'll have to keep a list of valid quote characters for each language, throw to whatever quote is still available (which will likely shred the previous variable escaping system), and throw an error if an Xpath consumes all valid quote characters for a language.

I will get this as well as I can for now. For XPaths, I will probably just not escape quotes at all and pencil that in for another day.

toddtarsi:
@zot - Alright, I concede. I'll merge your work as is. I thought I had it in me to do this super cleanly, but it is too much for my plate right now haha. I started looking at the selianize code and it does one thing, and the code-export does another thing, and if I'm going to get all into that mode, it's going to be about the v4 stuff. So I'm totally going to recommend we merge your code. If it solves your case and anyone else dealing with what you are, it looks good enough to me. I'm going to ask for someone else to give it another review.

I'm planning to deploy this week, and I'll do my best to have this in. That said, I don't have deploy control, so fingers crossed I can align the stars 😄

zot:
Hey, whatever works!

diemol (approved):

PR #1329, opened 2022-02-17
Fix stale frontend

toddtarsi:
Commits have relevant messaging

@marvelcoder - Please review, thanks mate 😄

### DIFF
--- a/packages/selenium-ide/src/browser/helpers/getActiveData.ts
+++ b/packages/selenium-ide/src/browser/helpers/getActiveData.ts
@@ -22,17 +22,11 @@ export const getActiveSuite = (session: CoreSessionData): SuiteShape => {
 export const getActiveTest = (session: CoreSessionData): TestShape => {
   const { project, state: { activeTestID } } = session
 
-  return useMemo(
-    () => project.tests.find(hasID(activeTestID)) || defaultTest,
-    [activeTestID, project.id]
-  )
+  return project.tests.find(hasID(activeTestID)) || defaultTest;
 }
 
 export const getActiveCommand = (session: CoreSessionData, activeTest?: TestShape): CommandShape => {
   const { state: { activeCommandID } } = session
   const _activeTest = activeTest || getActiveTest(session);
-  return useMemo(
-    () => _activeTest.commands.find(hasID(activeCommandID)) || defaultCommand,
-    [activeCommandID, _activeTest.id]
-  )
+  return _activeTest.commands.find(hasID(activeCommandID)) || defaultCommand;
 };
--- a/packages/selenium-ide/src/browser/windows/ProjectTestCommandList/components/CommandList.tsx
+++ b/packages/selenium-ide/src/browser/windows/ProjectTestCommandList/components/CommandList.tsx
@@ -6,7 +6,7 @@ import TableCell from '@material-ui/core/TableCell'
 import TableContainer from '@material-ui/core/TableContainer'
 import TableHead from '@material-ui/core/TableHead'
 import TableRow from '@material-ui/core/TableRow'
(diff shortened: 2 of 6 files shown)


packages/selenium-ide/src/browser/helpers/getActiveData.ts
```diff
@@ -22,17 +22,11 @@ export const getActiveSuite = (session: CoreSessionData): SuiteShape => {
 export const getActiveTest = (session: CoreSessionData): TestShape => {
   const { project, state: { activeTestID } } = session
 
-  return useMemo(
-    () => project.tests.find(hasID(activeTestID)) || defaultTest,
-    [activeTestID, project.id]
-  )
+  return project.tests.find(hasID(activeTestID)) || defaultTest;
```
MarvelCoder, line 25: @toddtarsi Can you tell me why this part was removed , [activeTestID, project.id] ?

packages/selenium-ide/src/browser/windows/ProjectTestCommandList/components/CommandList.tsx
```diff
@@ -6,7 +6,7 @@ import TableCell from '@material-ui/core/TableCell'
 import TableContainer from '@material-ui/core/TableContainer'
 import TableHead from '@material-ui/core/TableHead'
 import TableRow from '@material-ui/core/TableRow'
-import { Paper } from '@material-ui/core'
+import { Box, Divider, Paper, Typography } from '@material-ui/core'
```
MarvelCoder, line 9: @toddtarsi So I believe this is to include more UI element classes, to make data in row appear more sensible.

packages/selenium-ide/src/main/session/controllers/Projects/index.ts
```diff
@@ -28,19 +29,26 @@ export default class ProjectsController {
...
+        return window.isFocused()
+      }, false)
+      if (!anyWindowFocused) {
+        childWindows.forEach((win) => {
+          win.hide()
+        })
+      }
     })
-    mainWindow.on("closed", () => {
+    mainWindow.on('closed', () => {
```
MarvelCoder, line 51: @toddtarsi I noticed " " being replaced by ' 'in this whole file, is this followed by Typescript or it's a tab vs spaces argument  ?

toddtarsi: @MarvelCoder - This bit is due to the linting system(s). We use two linter configs that work together (and sometimes fight each other), eslint and prettier. This particular rule is I believe due to the singleQuote rule in the top level .prettierrc file. I don't think about it much, and just call the "Format Document" command in VSCode a lot whenever I'm working.

packages/selenium-ide/src/browser/windows/ProjectTestCommandList/components/CommandList.tsx
toddtarsi: @MarvelCoder - Yessir, it looks pretty ugly now, but I'd like to try to convert the horizontal real estate in the commands list to a more vertical real estate, to clear more space for the main playback / recording window.

packages/selenium-ide/src/browser/helpers/getActiveData.ts
toddtarsi: Yes. So, React.useMemo is an optimization in React. It basically says 'memoize the return value of this function every time these values change'. However, I didn't factor in that the memoization would cause changes to the individual test commands (like if a target or value changed for example) to not get propagated down. As a result, I'd rather just drop all memoization for now, and only add that back later in cases where we see a need for optimization.

MarvelCoder (approved):

PR #1349, opened 2022-03-22
slight linting and adding new project plugin edit commands

toddtarsi:
@coinzdude @MarvelCoder - This PR is a bit of an example of how to add new endpoints to the frontend. Please check if you guys have any questions. The mutator function is effectively something that mutates a common store between the frontend and backends. To keep things secure, we isolate everything except our API, which performs matching mutations between the main process and the renderer process. This is meant to keep things as safe as possible in the frontend. We further restrict this by only exposing the recorder api in playback windows.


coinzdude (approved):
I need a good bit more experience to provide good feedback here. I like the incorporation of @mui/material.

coinzdude (changes requested):
I encountered an issue building locally.

```
ERROR in ./src/main/session/controllers/Driver/index.ts 76:25-31
TS6133: 'driver' is declared but its value is never read.
    74 |     })
    75 |   }
  > 76 |   async onPlaybackStart(driver: WebDriverExecutor) {
       |                         ^^^^^^
    77 |     const playbackWindow = await this.session.windows.get('playback-window')
    78 |     // Figure out playback window from document.title
    79 |     if (!this.windowHandle) {

coinzdude:
yarn clean and force git and all builds fine.

coinzdude (approved):
Builds and runs

coinzdude (approved):
Builds and runs

toddtarsi:
@coinzdude - Thanks for the review!

PR #1356, opened 2022-03-30
Plugin handling and package.json cleanup

toddtarsi:
Does the following things:

1. Adds handlers for editing plugins
2. Consolidates plugin loaders to use side-runtime
3. Gets plugins working with side-cli
4. Gets side-cli working on the latest version of the dependencies
5. Makes the various child package.json work a bit more consistently
6. Removes all extraneous dependencies from the top level package.json
7. Consolidates the toolbar buttons to be a bit cleaner to re-use for various states
8. Fixes a bug that caused the stop button while recording to hang


coinzdude:
This builds for me

PR #1369, opened 2022-04-16
Swapping playback and project window focus status

coinzdude:
Having playback window own focus for the app was causing some interactivity and focus issues on Windows usage.


toddtarsi:
@coinzdude - This PR looks really good to me. Nice work!

PR #1373, opened 2022-04-17
Cross platform yarn start

coinzdude:
Need a little more control to perform os variations on yarn tasks.

Include real fix for windows process kill end of chromedriver.exe when exiting.  electron was calling 'cmd' which was causing the wrong pid to be tracked and killed on windows


toddtarsi:
This LGTM, nice work as always @coinzdude

PR #1374, opened 2022-04-19
Begin View menu, start with 'show DevTools'

coinzdude:
I took a stab at creating a View menu to have a shortcut key to show DevTools. Want to include this in other generated Windows soon


packages/selenium-ide/src/main/session/controllers/Menu/menus/projectView.ts
```diff
@@ -0,0 +1,23 @@
+import { BrowserWindow} from 'electron'
```
toddtarsi, line 1: @coinzdude - Can you hoist this up to menus/index.ts? This way, we can call it by name via API and popup the show dev tools thing on a right click in the playback windows.

PR #1378, opened 2022-04-21
 allow repeated loading of scripts, do not close project window. new menu, View, Playback

coinzdude:


packages/selenium-ide/src/main/session/controllers/Menu/menus/projectView.ts
```diff
@@ -1,16 +1,23 @@
...
 import { Menu } from 'electron'
 import { MenuComponent, Session } from 'main/types'
 
 export const projectViewCommands: MenuComponent = (session) => async () =>
   [
+    {
+      accelerator: 'CommandOrControl+P',
+      label: 'Show Playback Window',
+      click: async () => {
+        await session.windows.onPlaybackShow()
```
toddtarsi: @coinzdude - Nice use of menus!

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -158,34 +158,43 @@ export default class WindowsController {
...
-      const windows = BrowserWindow.getAllWindows()
-      const anyWindowFocused = windows.reduce((focused, window) => {
-        if (focused) return true
-        return window.isFocused()
-      }, false)
-      if (!anyWindowFocused) {
-        playbackWindow.hide()
-      }
-    })
+    projectWindow.on('focus', () => {})
```
toddtarsi: @coinzdude - Can you delete these event listeners if they're not doing anything?

coinzdude: remove empty event listeners

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -158,34 +158,43 @@ export default class WindowsController {
     )
   }
 
+  async onPlaybackShow() {
+    await this.open(playbackWindowName)
```
toddtarsi: @coinzdude - I think you want this to do this instead:

```
  async onPlaybackShow() {
    const window = await this.open(playbackWindowName)
    this.handlePlaybackWindow(window)
  }
```

That second function wires up the things for recording in subframes and such.

toddtarsi: Oh wait no I'm dumb, that's called in the sub function. My bad!

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -123,7 +123,7 @@ export default class WindowsController {
     }
     this.windows[name] = window
     window.on('closed', () => {
-      delete this.windows[name]
+      if (name != 'project-editor') delete this.windows[name]
```
toddtarsi: Good catch!

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: Hmmm.  I'm not certain actually.  Here's a code change for it and it seems to be fine?

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -158,34 +158,43 @@ export default class WindowsController {
...
     await this.open(projectEditorWindowName)
     await this.close('splash')
 
+    BrowserWindow.getAllWindows().forEach((win) => {
+      console.debug('found win, closable ' + win.closable)
+      console.debug(win.id)
+      console.debug(win.accessibleTitle)
+
+      // TODO implement an enum on the window shape to include "type" so that we aren't looking at strings.  Also, this name in the title is subject to change, for live URL or localization, etc.
+      if (win.title == 'Playback Window') {
```
toddtarsi: This is a total nitpick, but strong equality `===` is better for most strings I think

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
toddtarsi: @coinzdude - Oh! I just realized, this should probably be this:

```
if (name !== projectEditorName) delete this.windows[name]
```

toddtarsi (changes requested):
This PR looks great overall. I have a couple total nitpicks around strong equality, and only using soft equality when doing funky stuff with null or undefined kinda shapes, but thats trivial. The overall PR looks very good. Nice work!!

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
toddtarsi: Yeah, you're right. This is not a good change I requested. I wrote this before I realized the open function already called this. That was my bad. Good call! 😄

coinzdude:
Reverting the view playback menu

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: made stronger!

toddtarsi: Nice work!

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: Thanks for that. I ended up swapping to check for anything that does not begin with this string so we can name the window in the future like "Project Editor - My Script"

toddtarsi: Oh neat! Very cool

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -123,7 +123,8 @@ export default class WindowsController {
     }
     this.windows[name] = window
     window.on('closed', () => {
-      delete this.windows[name]
+      if (!name.startsWith(projectEditorWindowName))
```
toddtarsi: @coinzdude - What is going on with this line? I'm a bit worried about keeping a stale reference to a closed window.

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -159,33 +160,50 @@ export default class WindowsController {
...
+        https://github.com/SeleniumHQ/selenium-ide/blob/trunk/packages/selenium-ide/src/main/session/controllers/Windows/index.ts#L93
+      
+      implement an enum on the window shape to include "type" so that we aren't looking at strings.  Also, this name in the title is subject to change, for live URL or localization, etc.
+       */
+
+      if (!win.title.startsWith('Project Editor') && (!win.title.startsWith('Command List'))) {
+        win.closable = true
+        win.close()
+      }
+      this.playbackWindows.forEach((bw) => {
```
toddtarsi: @coinzdude - This part should probably be outside of the iterator going over each window.

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -159,33 +160,50 @@ export default class WindowsController {
...
-    })
     projectWindow.on('closed', () => {
+      console.debug('projectWindow on closed')
       BrowserWindow.getAllWindows().forEach((win) => {
-        if (!win.closable) win.closable = true
-        win.close()
+        console.debug('found win, closable ' + win.closable)
+        if (!win.closable) {
+          win.closable = true
+          win.close()
```
toddtarsi: @coinzdude - I think you'll want to call win.close() even if window is already closable

coinzdude: Great catch. I think I got this now?

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: I'm not really sure how this works, or the 'right' way to close these.  I don't know how to get these outside of the iterator either. Should I put the windows to close into another list and then close those?

coinzdude: I submitted some changes here. Let me know.

packages/selenium-ide/src/browser/windows/ProjectEditor/components/Suites/SuitesDrawer.tsx
```diff
@@ -49,16 +49,27 @@ const SuiteList: FC<SuiteListProps> = ({
...
-          name={name}
-          rename={rename}
-          selected={id === activeSuite}
-          setSelected={setSelected}
-        />
-      ))}
+      {suites
+        .slice()
+        .sort((a, b) => {
+          if (b.name < a.name) {
```
toddtarsi: @coinzdude - Okay, one more critique here. I'd recommend changing this to use localeCompare:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

```
.sort((a, b) => a.localeCompare(b))
```

It will reduce the complexity of this and alphabetize things in a more cross-language-y way.

packages/selenium-ide/src/browser/windows/ProjectEditor/components/Tests/TestsDrawer.tsx
```diff
@@ -42,16 +42,27 @@ const TestList: FC<TestListProps> = ({ activeTest, open, setOpen, tests }) => (
...
-          key={id}
-          name={name}
-          rename={rename}
-          selected={id === activeTest}
-          setSelected={setSelected}
-        />
-      ))}
+      {tests
+        .slice()
+        .sort((a, b) => {
```
toddtarsi: Same as above here

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -159,32 +159,53 @@ export default class WindowsController {
...
+       also, look at this idea:
+       We should delete all the windows we can when we call the projects.close endpoint, which is part of projects.load, but should probably also be on projects.new. This function here does the cleanup and is rather redundant and somewhat confused, much like myself lol:
+        https://github.com/SeleniumHQ/selenium-ide/blob/trunk/packages/selenium-ide/src/main/session/controllers/Windows/index.ts#L93
+      
+      implement an enum on the window shape to include "type" so that we aren't looking at strings.  Also, this name in the title is subject to change, for live URL or localization, etc.
+       */
+
+      if (
+        !win.title.startsWith('Project Editor') &&
+        !win.title.startsWith('Command List')
```
toddtarsi, line 188: I don't think this will ever be a title we need to worry about, so this line can be dropped.

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -159,32 +159,53 @@ export default class WindowsController {
   }
 
   async onProjectLoaded() {
-    await this.close(projectEditorWindowName)
+    // await this.close(projectEditorWindowName)
```
toddtarsi: You can drop this line instead of commenting it. It can go. I can see that we aren't closing the project window now.

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
toddtarsi: Honestly, you might be able to get rid of this entire block. The only windows that should be open after forcing closed splash should be the project editor window and N playback windows. By iterating those, I don't think this block should run for any windows.

toddtarsi: To be clear, I mean the whole `BrowserWindow.getAllWindows().forEach(...)` block

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -159,32 +159,53 @@ export default class WindowsController {
...
-      const anyWindowFocused = windows.reduce((focused, window) => {
-        if (focused) return true
-        return window.isFocused()
-      }, false)
-      if (!anyWindowFocused) {
-        playbackWindow.hide()
-      }
-    })
     projectWindow.on('closed', () => {
+      console.debug('projectWindow on closed')
```
toddtarsi, line 202: I think this logger can go, same with the logger two lines down. If you want to keep them, we might want to move them behind a namespace like we do here:

https://github.com/SeleniumHQ/selenium-ide/blob/11389d058c9850004d34c165b5c7aa2b84225645/packages/selenium-ide/src/main/api/classes/Handler.ts#L7

packages/selenium-ide/src/browser/windows/ProjectEditor/components/Suites/SuitesDrawer.tsx
coinzdude: Much better!

packages/selenium-ide/src/browser/windows/ProjectEditor/components/Tests/TestsDrawer.tsx
coinzdude: done!

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: I tried ignoring 'command list' but I couldn't as it would get closed.  'command list' appears to be a temporary name on the way to getting the name 'project editor'

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: done

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: On this, in my testing, if I don't exclude both of these window names, then the app will close when loading a project

toddtarsi (approved):
LGTM 👍 Nice work!

toddtarsi (approved):
Nice work. Merging!

PR #1396, opened 2022-04-26
new View, Reset Playback Windows, check if PB window needed

coinzdude:
Hot key to remove all playback windows and reset to one. Also, instantiate a playback window if all are closed on play.


packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -187,18 +187,29 @@ export default class WindowsController {
...
+      if (!bw.closable) {
+        bw.closable = true
+      }
+      bw.close()
+    })
+
+    await this.close(playbackWindowName)
+    await this.open(playbackWindowName)
+    const playbackWindow = await this.get(playbackWindowName)
+    this.handlePlaybackWindow(playbackWindow)
```
toddtarsi: @coinzdude - This line is called as part of the open command if the name matches 'playback-window', so this line can go.

toddtarsi (approved):
LGTM 👍 Nice work!

PR #1401, opened 2022-05-02
Check for save on load/exit; partially working

coinzdude:
Interested in your help with this one. It works for loading new files, but cancel button and app close do not yet work


toddtarsi (approved):
LGTM 👍 Nice work!

PR #1402, opened 2022-05-02
Save and quit, allow for cancellation

toddtarsi:
Exposes a boolean within checkIfCurrentProjectChanged

If truthy, continue the action.
If falsy, abort.

Make load optionally handle null as its result shape instead of a new project
Expose quit as a manually instrumented menu function

Delete a commented out function


coinzdude (approved):
This all looks good to me!

PR #1405, opened 2022-05-02
Cleanup play

coinzdude:
reset playback windows on new play

PR #1406, opened 2022-05-02
Win quit fix

coinzdude:
Fix for last commit to allow resume and single command playback


toddtarsi:
This is the same code as https://github.com/SeleniumHQ/selenium-ide/pull/1405, but with a slightly different handler on play

PR #1408, opened 2022-05-03
Persist window

coinzdude:
Everything is working pretty well here, but I can't get the 'default' to return a value and I find I have to have a hard coded check for null and set a default.

Defaults are not returning for me here:
```
    size = storage.get<'windowSize'>('windowSize', [552, 664])
    position = storage.get<'windowPosition'>('windowPosition', [982, 20])
    if (!size[0]) size = [552, 664]
    if (!position[0]) position = [982, 20]
```


toddtarsi:
This feature is definitely in

PR #1411, opened 2022-05-04
Preserve project state

coinzdude:
Updated code based on PR feedback. I think this looks good.  Lightly tested.

### DIFF
--- a/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/CommandSelector.tsx
+++ b/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/CommandSelector.tsx
@@ -1,12 +1,10 @@
 import { CodeOff, HelpCenter } from '@mui/icons-material'
-import {
-  FormControl,
-  IconButton,
-  InputLabel,
-  MenuItem,
-  Select,
-  Tooltip,
-} from '@mui/material'
+import { FormControl } from '@mui/material'
+import { IconButton } from '@mui/material'
+import { InputLabel } from '@mui/material'
+import { MenuItem } from '@mui/material'
+import { Select } from '@mui/material'
+import { Tooltip } from '@mui/material'
 import React, { FC, useMemo } from 'react'
 import { setField, updateField } from './utils'
 import { CommandSelectorProps } from '../types'
--- a/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
+++ b/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
@@ -1,10 +1,10 @@
-import { HelpCenter } from '@mui/icons-material'
+import HelpCenter from '@mui/icons-material/HelpCenter'
 import AddToHomeScreenIcon from '@mui/icons-material/AddToHomeScreen'
-import { Autocomplete } from '@mui/material'
-import { FormControl } from '@mui/material'
-import { IconButton } from '@mui/material'
-import { TextField } from '@mui/material'
-import { Tooltip } from '@mui/material'
+import Autocomplete from '@mui/material/Autocomplete'
+import FormControl from '@mui/material/FormControl'
+import IconButton from '@mui/material/IconButton'
+import TextField from '@mui/material/TextField'
+import Tooltip from '@mui/material/Tooltip'
 import capitalize from 'lodash/fp/capitalize'
 import React, { FC, useEffect } from 'react'
 import { updateField, updateFieldAutoComplete } from './utils'
@@ -60,11 +60,7 @@ const CommandLocatorField: FC<CommandArgFieldProps> = ({
         onInputChange={onChangeAutoComplete}
         options={(command[fieldNames] ?? []).map((entry) => entry.join('='))}
         renderInput={(params) => (
-          <TextField
-            {...params}
-            label={label}
-            name={fieldName}
-          />
+          <TextField {...params} label={label} name={fieldName} />
         )}
         size="small"
         text-overflow="ellipsis"
@@ -78,11 +74,7 @@ const CommandLocatorField: FC<CommandArgFieldProps> = ({
       >
         <AddToHomeScreenIcon />
       </IconButton>
-      <Tooltip
-        className="mx-2 my-auto"
-        title={fullnote}
-        placement="top-end"
-      >
+      <Tooltip className="mx-2 my-auto" title={fullnote} placement="top-end">
         <HelpCenter />
       </Tooltip>
     </FormControl>
--- a/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
+++ b/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
@@ -21,7 +21,8 @@ const CommandLocatorField: FC<CommandArgFieldProps> = ({
   const fieldNames = (fieldName + 's') as PluralField
   const FieldName = capitalize(fieldName)
 
-  const note = commands[command.command][fieldName]?.description ?? ''
+  const fullnote = commands[command.command][fieldName]?.description ?? ''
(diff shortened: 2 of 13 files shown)


packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/CommandSelector.tsx
```diff
@@ -1,12 +1,10 @@
 import { CodeOff, HelpCenter } from '@mui/icons-material'
-import {
-  FormControl,
-  IconButton,
-  InputLabel,
-  MenuItem,
-  Select,
-  Tooltip,
-} from '@mui/material'
+import { FormControl } from '@mui/material'
```
toddtarsi, line 2: @coinzdude - At some point in here, I'd like to change the format of the import to more like this:

```
import FormControl from '@mui/material/FormControl'
```

Basically, this is one of the biggest continuous things you can do in JS to reduce your bundle size and optimize performance. By going more specific with your import, you bring in less, and the total package gets smaller and faster at executing.

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
```diff
@@ -1,9 +1,10 @@
+import { HelpCenter } from '@mui/icons-material'
```
toddtarsi: You especially want to tree shake with icon packages. One mistaken import like this, and you can add 5 megabytes onto your JS bundle. Nearly every popular icon package has issue reports around this stuff, and the fix is always the same.  Very good to go with this shape here:

```
import HelpCenter from '@mui/icons-material/HelpCenter'
```

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/TextField.tsx
```diff
@@ -1,10 +1,11 @@
-import { FormHelperText } from '@mui/material'
 import FormControl from '@mui/material/FormControl'
+import { HelpCenter } from '@mui/icons-material'
```
toddtarsi: 👊 🐴 💀 blah blah tree shaking

packages/selenium-ide/src/main/session/controllers/Projects/index.ts
```diff
@@ -103,12 +107,28 @@ export default class ProjectsController {
...
-    if (project) {
-      this.onProjectLoaded(project, filepath)
+  async load(filepath: string): Promise<CoreSessionData | null> {
+    const projectStates = storage.get<'projectStates'>('projectStates')
+    let loadedState: StateShape = {} as StateShape
+
+    if (projectStates[filepath]) {
+      loadedState = projectStates[filepath]
+      if (!loadedState.activeCommandID)
+        loadedState = clone(defaultState)
```
toddtarsi: Nice work having this in here. I think you can actually assign this to defaultState without the clone operator here. The reason I believe that is that the mutators all perform immutable updates, so the original object will never be modified by the mutators as they work. There might be something I'm missing in the controller ofc, but just sayin haha.

toddtarsi: @coinzdude - Also, I was wondering if we should use the project id field over file path, so you can move the file on your system, and still get all of the cached goodness

packages/selenium-ide/src/main/session/controllers/State/index.ts
```diff
@@ -19,7 +21,13 @@ export default class StateController {
   }
 
   async onProjectUnloaded() {
+    let projectStates = storage.get<'projectStates'>('projectStates')
+    if (this.session.projects.filepath) {
+      projectStates[this.session.projects.filepath] = this.state
+        storage.set<'projectStates'>('projectStates', projectStates)
```
toddtarsi: There's a teeny bit of funky indentation here. None of this is a deal breaker, I'm still merging btw. I just want to call out these little things.

Also, this sort of read, write flow isn't needed. electron-store actually supports nested key operations, so you could just do this:|

```
storage.set(`projectStates.${this.session.projects.filepath}`, this.state)
```

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -175,6 +178,17 @@ export default class WindowsController {
     this.handlePlaybackWindow(playbackWindow)
 
     const projectWindow = await this.get(projectEditorWindowName)
+    let position, size: number[]
+
+    size = storage.get<'windowSize'>('windowSize', [552, 664])
+    position = storage.get<'windowPosition'>('windowPosition', [982, 20])
+    if (!size[0]) size = [552, 664]
```
toddtarsi: These defaults shouldn't be needed because you're already assigning defaults with the get operation.

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -175,6 +178,17 @@ export default class WindowsController {
...
+    if (!position[0]) position = [982, 20]
+
+    projectWindow.setSize(size[0], size[1], true)
```
toddtarsi, line 186: Nice use of the window APIs!

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
coinzdude: I think some of these I'm just inheriting as already being a particular way in the files.  I'll submit this change for the material references.  Does this look good?

```
import HelpCenter from '@mui/icons-material/HelpCenter'
import AddToHomeScreenIcon from '@mui/icons-material/AddToHomeScreen'
import Autocomplete from '@mui/material/Autocomplete'
import FormControl from '@mui/material/FormControl'
import IconButton from '@mui/material/IconButton'
import TextField from '@mui/material/TextField'
import Tooltip from '@mui/material/Tooltip'
import capitalize from 'lodash/fp/capitalize'
import React, { FC, useEffect } from 'react'
import { updateField, updateFieldAutoComplete } from './utils'
import { CommandArgFieldProps } from '../types'
```

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/TextField.tsx
coinzdude: Is this good for this file?

```
import FormControl from '@mui/material/FormControl'
import HelpCenter from '@mui/icons-material/HelpCenter'
import TextField from 'browser/components/UncontrolledTextField'
import capitalize from 'lodash/fp/capitalize'
import React, { FC } from 'react'
import { CommandFieldProps } from '../types'
import { updateField } from './utils'
import Tooltip from '@mui/material/Tooltip'
```

packages/selenium-ide/src/main/session/controllers/Projects/index.ts
coinzdude: Much better! Will submit

packages/selenium-ide/src/main/session/controllers/State/index.ts
coinzdude: formatted.  On windows, I had to turn off the "format on save" because of that "last line can't be blank" issue we found

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
coinzdude: Going to try this code without the if check. For some reason I was not getting back my defaults.

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/LocatorField.tsx
toddtarsi: These look great!

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/TextField.tsx
toddtarsi: Perfectimundo!

PR #1412, opened 2022-05-04
Save and resume project state on reload is working in this version!

coinzdude:
Some very initial code to add fallback support to doClick


coinzdude:
use rimraf for 'yarn clean'

coinzdude:
Added LF mode for VS Code project settings to better support Windows development collaboration

PR #1414, opened 2022-05-09
Select frame matching fix

coinzdude:
This is a fix using a solution selectFrame where we had similar matching problems using C# and selenium. Perhaps there's a better explantation for what is happening and why this works, but this work-around seems to get the job done.


toddtarsi (approved):
LGTM 👍 Nice work, very clean. Thanks for putting something down to deal with the selectFrame issues for now

PR #1420, opened 2022-05-10
Handle bad files

coinzdude:
invalid files can get into the recents and can't be loaded. Added try/catch to handle ignoring them.


toddtarsi (approved):
LGTM 👍

packages/selenium-ide/src/main/session/controllers/Projects/index.ts
```diff
@@ -162,11 +162,18 @@ export default class ProjectsController {
...
+    let project: ProjectShape
+    try {
+      project = JSON.parse(fileContents)
+
+      project.plugins = project.plugins.filter(
+        (plugin) => typeof plugin === 'string'
+      )
+      return project
+    } catch (e) {
+      console.log((e as Error).message)
```
toddtarsi, line 174: @coinzdude - For messages like this, you might want to log using `console.error`. console.error, unlike console.log, produces a stack trace for whatever is reported

toddtarsi (approved):
Clean and well executed. Nice!

PR #1422, opened 2022-05-12
Feat: playback stabilizations

toddtarsi:
Make playback smoothly handle missing windows
Moved onBeforePlay out of driver and into playback
Made some types in side-runtime a bit more consistent
Move retry logic out of onClick event and into CommandNode.execute


coinzdude:
Looks great

PR #1425, opened 2022-05-17
New command pref attempt

coinzdude:
rough attempt to start user prefs outside of projectstate


packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
```diff
@@ -0,0 +1,34 @@
...
+  update(
+    'state.userprefs',
+    () =>
+      setUserPrefs(userPrefs),
+    session
+  )
+
+export const browser = browserHandler<Shape>()
+
+export const main = mainHandler<Shape>(passthrough)
```
toddtarsi: @coinzdude - Remove passthrough from here. This tells the server basically to immediately mutate and emit the onMutate event. In this case, you want to do some storage stuff, so it isn't a passthrough

packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const setUserPrefs = (prefs: UserPrefs) => {
```
toddtarsi: @coinzdude - This needs to not be here. Instead, make a method called toggleUserPref on StateController that does what you want.

packages/selenium-ide/src/api/helpers/userPrefs.ts
```diff
@@ -0,0 +1,15 @@
+import { UserPrefs } from 'api/types'
+import storage from 'main/store'
+
+export const getUserPrefs = (): UserPrefs => {
```
toddtarsi: @coinzdude - This file needs to go

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
```diff
@@ -57,6 +61,24 @@ const ProjectTab: FC<ProjectTabProps> = ({ session: { project } }) => (
...
+          <InputLabel id="insertNewCommandPref">
+            New command insert preference
+          </InputLabel>
+          <Select
+            id="insertNewCommandPref"
+            label="Insert new command placement preference"
+            name="insertNewCommandPref"
+            value={'After'}
+            onChange={(e: any) => {
+              // toggleUserPref({insertCommandPref: e.target.value})
```
toddtarsi: This should be `window.sideAPI.state.toggleUserPref`

coinzdude: Ok, think I got this.

packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
coinzdude: I think I got this

coinzdude: I had missed the unused passthrough which was caught on compile.  Fixed.

packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
```diff
@@ -0,0 +1,18 @@
+import browserHandler from 'browser/api/classes/Handler'
+import update from 'lodash/fp/update'
+import mainHandler from 'main/api/classes/Handler'
+import { Mutator } from 'api/types'
+import {Session} from 'main/types'
+
+export type Shape = Session['state']['toggleUserPref'];
+
+export const mutator: Mutator<Shape> = (session) =>
```
toddtarsi: export const mutator: Mutator<Shape> = (session, { params: [newPrefValue] }) =>

packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
```diff
@@ -0,0 +1,18 @@
...
+  update(
+    'state.insertCommandPref',
+    (pref) => (pref === 'Before' ? 'After' : 'Before'),
```
toddtarsi: (pref) => newPrefValue,

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
```diff
@@ -57,6 +61,23 @@ const ProjectTab: FC<ProjectTabProps> = ({ session: { project } }) => (
...
+          <InputLabel id="insertNewCommandPref">
+            New command insert preference
+          </InputLabel>
+          <Select
+            id="insertNewCommandPref"
+            label="Insert new command placement preference"
+            name="insertNewCommandPref"
+            value={state.userPrefs.insertCommandPref}
+            onChange={(e: any) => {
+              window.sideAPI.state.toggleUserPref(e.target.value)
```
toddtarsi: This is being called with only a string on the browser side, and then you've got a controller function on the other side with some object type. Your linter should not be okay with one side of this. It's a very clear disagreement of types.

packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
coinzdude: Moved

packages/selenium-ide/src/api/commands/state/toggleUserPref.ts
coinzdude: This one is confusing to me, but I did implement as an update.

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
coinzdude: I'm not seeing any linter complaint, but it's also not yet working

packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -170,6 +170,10 @@ export default class WindowsController extends BaseController {
     const projectWindow = await this.get(projectEditorWindowName)
     const size = storage.get<'windowSize'>('windowSize')
     const position = storage.get<'windowPosition'>('windowPosition')
+
+    // const userPrefs = storage.get<'insertCommandPref'>('insertCommandPref')
```
toddtarsi: @coinzdude - The changes in this file should be dropped.

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
```diff
@@ -57,6 +61,23 @@ const ProjectTab: FC<ProjectTabProps> = ({ session: { project } }) => (
...
+          <Select
+            id="insertNewCommandPref"
+            label="Insert new command placement preference"
+            name="insertNewCommandPref"
+            value={state.userPrefs.insertCommandPref}
+            onChange={(e: any) => {
+              window.sideAPI.state.toggleUserPrefInsert(e.target.value)
+            }}
+          >
+            <MenuItem value={'After'}>After</MenuItem>
```
toddtarsi: You can simplify static strings like this in react. I'm not sure if they currently do but I think they can optimize for this case in some way.

```
<MenuItem value="After">
```

packages/selenium-ide/src/api/commands/state/toggleUserPrefInsert.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Session } from 'main/types'
+
+export type Shape = Session['state']['toggleUserPrefInsert']
+
+export const mutator: Mutator<Shape> = (session, { params: [insertCommandPref] }) =>
+  update(
+    'state.userprefs.insertCommandPref',
+    (userPrefs: UserPrefs) =>
+      userPrefs.insertCommandPref = insertCommandPref
+      ,
```
toddtarsi: I believe lodash fp does "immutable" type operations. They are designed to not mutate the original shape but return a new shape. You know, so that way we can make obtuse reduce functions and brick things on shit like simple counters. In a case where you're just setting a nested key, lodash/fp/set is probably a better choice than lodash/fp/update:

```
import set from 'lodash/fp/set'

export const mutator: Mutator<Shape> = (session, { params: [insertCommandPref] }) => set(
  'state.userprefs.insertCommandPref',
  insertCommandPref,
  session,
);
```

Something like this should work ^^^

toddtarsi:
Okay, those are all of my critiques about the code for now. They're largely pedantic, and meant to be half helpful and half just purely anxious about code smell about little things and such. All and all, this feature is shaping up well. Nice work!
I actually really like your most recent commit messages in git. They're small and frustrated, but they remind me of things I type for commit messages when I'm in serious iteration mode. I do still think you want to get in the practice of having minimal commits in a PR. I think the "ideal" currently is a PR as 1 commit with basically a couple paragraphs talking about it.

packages/selenium-ide/src/api/commands/state/toggleUserPrefInsert.ts
coinzdude: almost following, ha!  thanks!

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
coinzdude: Good to have, thanks

coinzdude: I tried that and the drawn select list options were blank. I didn't find a better or more succinct way and only see others using some kind of onclick handler and then processing the value out of the passed event. That seems like a bit of code overhead to just get what was selected.  

Here's a mega thread with the bundles of options on writing click handlers. Some of the examples here are wow. 
 https://stackoverflow.com/questions/43870814/get-value-of-menuitem-material-ui

coinzdude:
I made the change to use 'set' instead of change in the mutator, and it seems like the code is running which updates the state, but I don't yet see that the user selection impacts the prefs tab for the select list.

packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
toddtarsi: Do you mean the quoting style won't work? I know we use this component without issue here:

https://github.com/SeleniumHQ/selenium-ide/blob/eaf634a3728ed756344a8969b7cb0cdc5848f3a0/packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Tests/CommandFields/CommandSelector.tsx#L49

toddtarsi: Eh, we do this quoting style elsewhere unfortunately. Likely this doesn't have anything to do with it. My bet would be that you need those mutator changes to properly update the session shape.

coinzdude: I have this using the simpler syntax, and in debugging, I can see that the state.userPrefs.insertCommandPref is changing and being set, but the UI is not getting or using the newer updated state info. It's stuck, seemingly hardcoded to 'After'

coinzdude:
Last bit of code submitted now to load InsertCommandPref and userPrefs on project load and to consider this value when adding commands

PR #1432, opened 2022-05-30
New command pref

coinzdude:
rewrote insert check as conditional


toddtarsi (approved):
@coinzdude - This looks great, thanks for getting those other things!

PR #1433, opened 2022-05-31
Autocomplete command

coinzdude:
This renders the select list well, and supports the auto-complete as wanted, but I'm unable to figure out how to properly process the below which works well when a select list.  I think related to the autocomplete items being objects of  { label: item.name, id: item.id } but I'm struggling to debug this area.  

Definitely appreciate the expertise to learn what I am missing here, either to debug better, or to just know what is off base.

```typescript
         onChange={updateCommand(testID, command.id)}
```


coinzdude:
Autocomplete is working with this commit for command editor.

image

toddtarsi:
Closing, work was carried along elsewhere

PR #1434, opened 2022-05-31
Corrupt side file guard

coinzdude:
I added this check for a corrupt side file, which was breaking the IDE. 
Maybe this should go in the loader, though, as a general check?  Let me know if you know a better approach.

In my case I had a number in the command text property

```
        {
          "id": "8d05890c-74be-4f36-a3bb-a77a6719ce1b",
          "comment": "",
          "command": 0,
          "target": "alert('hi 1')",
          "targets": [],
          "value": ""
        },
```


toddtarsi (approved):
Merging

toddtarsi:
LGTM 👍 Nice guards

PR #1439, opened 2022-06-02
Build tweak

coinzdude:
I received this recommendation from VS code to improve time to set/hit breakpoints (I think it was).  Let me know what you think.


toddtarsi (approved):
This lgtm. Nice work!

PR #1441, opened 2022-06-02
New command pref

coinzdude:
Fix recording when pref set to Before


toddtarsi:
Closing, merged elsewhere

PR #1443, opened 2022-06-03
Fix delete first command

coinzdude:
fixes crash when first command in test deleted.

Thanks for the help to get this working!


toddtarsi (approved):
LGTM 👍

PR #1446, opened 2022-06-04
Playtohere fix

coinzdude:
Code was not handling with commandID was a CommandShape object


packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
```diff
@@ -1,4 +1,5 @@
 import CloseIcon from '@mui/icons-material/Close'
+import { FormHelperText } from '@mui/material'
```
toddtarsi: @coinzdude - Nitpick, but this should use naming pattern from below.

coinzdude: done

packages/selenium-ide/src/main/session/controllers/Menu/menus/testEditor.ts
```diff
@@ -100,60 +104,57 @@ export const recorderList = (session: Session) => async () => {
...
-        await session.api.playback.play(sessionData.state.activeTestID, [
-          0,
-          activeTest.commands.findIndex(
-            (cmd) => cmd.id === commandID
-          ),
-        ])
+export const playbackList: MenuComponent =
+  (session) => async (_commandID?: string) => {
+    const sessionData = await session.state.get()
+    const commandID = _commandID || getActiveCommand(sessionData)
```
toddtarsi, line 110: Oh, wow, I see! I'm using the wrong shape for when _commandID is falsy. Nice catch! Could you move this fix up to line 110?

```
  const commandID: string = _commandID || getActiveCommand(sessionData).id
```

PR #1450, opened 2022-06-06
Stop fix and play after p2h

coinzdude:
There's still an issue that when hitting play, after stop, after "play to here" the script will pause on the first line on the first try


toddtarsi:
Closing this too, this code got merged elsewhere. It's in trunk.

PR #1453, opened 2022-06-08
Fix delete first command

coinzdude:
re-apply after code shift


toddtarsi (approved):
Thanks, sorry again about that!

PR #1454, opened 2022-06-08
Fix delete first command

coinzdude:
crlf keeps switching back, changing to lf again here


toddtarsi:
Closing, this has dual curlies in the one file and the other file is already applied.

PR #1455, opened 2022-06-08
P2h stop play hack

coinzdude:
reapply the p2h stop/play fix-hack


packages/side-api/src/commands/tests/removeSteps.ts
```diff
@@ -24,14 +24,16 @@ export const mutator: Mutator<Shape> = (
...
   return update(
     `state.editor.selectedCommandIndexes`,
-    (selectedCommandIndexes: number[]) => {
-      const filteredCommands = selectedCommandIndexes.filter(
-        (_id, index) => !stepIndexes.includes(index)
-      )
-      if (filteredCommands.length) {
-        return filteredCommands
+    (selectedCommandIndexes: number[]): number[] => {
+      {
```
toddtarsi, line 28: @coinzdude - I think theres a syntax thing here. I see two curlies in a row, which seems wrong to me.

toddtarsi:
Ah, actually, closing this. We resolved this via adding something else I can't remember what.

PR #1457, opened 2022-06-09
Reapply newcommand pref round 2

coinzdude:
the manual insert and recording insert are working well with this


coinzdude:
Closing this PR as it includes a huge swath of crlf file changes that should not be here

PR #1460, opened 2022-06-09
sort suites by name

coinzdude:
From Suite list


coinzdude:
my mistake, closing again

PR #1476, opened 2022-06-27
Camel case pref

coinzdude:
new user pref to determine usage of forced camel casing of various content texts, primarily the list of tests in the suites project tab


toddtarsi:
Okay, merging this for now. Some point in the future, I'd like to change this from a boolean to an enum, but we can get there when we need to.

toddtarsi (approved):
LGTM 👍

PR #1477, opened 2022-06-27
Fix arrow key for command selector

coinzdude:
arrow keys were changing selected commands while using autocomplete in the command selector. This skips changing commands if auto-complete for command is the active element.


toddtarsi:
closing this, code is duplicated here:

https://github.com/SeleniumHQ/selenium-ide/pull/1480/files?w=1

toddtarsi:
Oh no wait I'm going to merge forward in history with squashes. That'll probably be cleanest

toddtarsi:
nm I tried to rebase and its too messy so I'm just closing and using latest to reduce duplicate code reviewing

PR #1480, opened 2022-06-29
Load recent

coinzdude:
New File menu option in project editor to load recent, which loads the splash page. I wanted the current project window to go away before the splash page loads, but that then keeps the project window from opening after selecting from splash, and it's not that disruptive to leave the current project editor temporarily in the background. This seems to work fine.


packages/selenium-ide/src/browser/windows/ProjectEditor/tabs/Project/ProjectTab.tsx
```diff
@@ -95,6 +95,23 @@ const ProjectTab: FC<ProjectTabProps> = ({ session: { project, state } }) => (
...
+        <FormControl>
+          <InputLabel id="camelCaseNames">
+            Camel case various names in UI
+          </InputLabel>
+          <Select
+            id="camelCaseNamesPref"
+            label="Camel case various names in UI"
+            name="camelCaseNamesPref"
+            value={state.userPrefs.camelCaseNamesPref}
+            onChange={(e: any) => {
```
toddtarsi, line 107: All and all this PR looks really clean from code.
A small critique I have here is I'd recommend doing the handler this way:

```
            onChange={(e) => {
              window.sideAPI.state.toggleUserPrefCamelCase(e?.target?.value ?? '')
            }}
```

React can intuit that e is something like `React.ChangeEventHandler<HTMLSelectElement>` or something like that. This instead makes sure the type provided is a string at function boundary. I'm going to squash merge because the history has overlap, but all and all this looks good. Nice work!

packages/side-api/src/models/state/index.ts
```diff
@@ -33,15 +33,18 @@ export const defaultRecorderState: RecorderStateShape = {
 
 export type InsertCommandPref = 'Before' | 'After'
 export type ThemePref = 'Dark' | 'Light' | 'System'
+export type CamelCaseNamesPref = 'Yes' | 'No'
```
toddtarsi, line 36: @coinzdude - I feel like this pref might be better generalized so its more easily extended later.

```
export type CaseNamesPref = 'caml' | 'identity'
```

toddtarsi (approved):
Code is good, squashing and merging

toddtarsi:
@coinzdude - Nice work on this code. This is a great batch of fixes! Thanks as always!

PR #1484, opened 2022-07-01
Dialog prompt for new test name

coinzdude:
This is working to create a new test name, but I can't work out getting TestNewDialog to show on subsequent "Adds"
@toddtarsi, for us to look at


coinzdude:
LGTM, 
This worked great, thanks for your coaching @toddtarsi

PR #1492, opened 2022-07-11
[WIP] Work towards in-IDE-Code-Export

toddtarsi:
This does the following:

1. Adds JS processing to build pipeline to do code export libraries
2. Begins typing csharp commons as an experiment in the cost of typing a code export library. Also so that I know that plugin devs can use valid types to guide them on this stuff.
3. Lays the groundwork for a 'try' command, to allow for trying blocks without failing a test
4. Adds a controller around code exporting to handle this stuff.

There's a lot more to do here, but its a start


toddtarsi:
Closing this, another PR goes further

PR #1493, opened 2022-07-12
Fix command target autocomplete selection

coinzdude:
clean up autocomplete events to allow selection of targets in command editor

### DIFF
--- a/packages/selenium-ide/src/main/session/controllers/Windows/index.ts
+++ b/packages/selenium-ide/src/main/session/controllers/Windows/index.ts
@@ -186,7 +186,7 @@ export default class WindowsController extends BaseController {
 
     if (size.length) projectWindow.setSize(size[0], size[1], true)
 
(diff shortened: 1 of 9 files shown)


packages/selenium-ide/src/main/session/controllers/Windows/index.ts
```diff
@@ -181,7 +185,26 @@ export default class WindowsController extends BaseController {
     const position = storage.get<'windowPosition'>('windowPosition')
 
     if (size.length) projectWindow.setSize(size[0], size[1], true)
-    if (position.length) projectWindow.setPosition(position[0], position[1])
+
+    var screenElectron = electron.screen.getPrimaryDisplay()
```
toddtarsi: @coinzdude - This should be a const

coinzdude: changed, good catch

toddtarsi (approved):
LGTM 👍

PR #1494, opened 2022-07-13
Full vscode debug

coinzdude:
tweaks to launch.json and new compound mode to allow debugging of front-end and node side of electron app with 
"Electron: All"


toddtarsi (approved):
LGTM 👍

PR #1495, opened 2022-07-13
Var to const

coinzdude:
I think this was lost in the squash


toddtarsi:
Work is done in another branch, closing this one

PR #1496, opened 2022-07-14
User pref to ignore SSL errors

coinzdude:
script developers may need to ignore SSL errors, e.g. when developing against some local instance, or explicitly not ignore SSL errors, e.g. production validation.

This gives the user the choice.


toddtarsi:
Closing, merged as part of another PR

PR #1498, opened 2022-07-14
Fix locators display

coinzdude:
the select list was displaying a join of the locator and the locator type which makes editing difficult.  This now displays only the locator without the method info.


toddtarsi (approved):
This LGTM 👍 , merging!

@coinzdude - Nice work!

PR #1499, opened 2022-07-15
Fix selectframe

coinzdude:
Split select frame recording activities into separate commands as needed to span multiple frame hops

### DIFF
--- a/packages/selenium-ide/src/main/session/controllers/Recorder/index.ts
+++ b/packages/selenium-ide/src/main/session/controllers/Recorder/index.ts
@@ -83,11 +83,14 @@ export default class RecorderController extends BaseController {
   async getFrameLocation(event: Electron.IpcMainEvent): Promise<string> {
     let frameLocation = 'root'
     let activeFrame = event.senderFrame
+    let pathParts = []
     while (activeFrame.parent) {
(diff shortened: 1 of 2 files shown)


packages/selenium-ide/src/main/session/controllers/Recorder/index.ts
```diff
@@ -83,11 +83,14 @@ export default class RecorderController extends BaseController {
...
     let activeFrame = event.senderFrame
+    let pathParts = []
     while (activeFrame.parent) {
       const frameIndex = activeFrame.parent.frames.indexOf(activeFrame)
-      frameLocation += `/${frameIndex}`
+      pathParts.push(frameIndex)
       activeFrame = activeFrame.parent
     }
+    pathParts.push(frameLocation)
+    frameLocation = pathParts.reverse().join('/')
```
toddtarsi, line 102: @coinzdude - I see that the reverse here would technically make the frame list order backwards from what it was previously. Did I have it backwards or something?

coinzdude: This is correct now because the frame order is built originating from the source element, so the frame relative paths need to be reversed before constructed into a path string

toddtarsi (approved):
LGTM 👍 Nice work @coinzdude

PR #1500, opened 2022-07-15
Fix selectframe

coinzdude:
This implements recording and playback for nested iframe conditions and matches on index


toddtarsi (approved):
LGTM 👍 Nice work!

PR #1503, opened 2022-07-19
Fix selectframe

coinzdude:
Adding this iframe URL matching fix back until we can get a better idea of why the indexes don't match on record vs playback, or until we can implement a new method to selectFrame by URL vs index.


toddtarsi:
Closing, merged as part of another PR

PR #1507, opened 2022-07-25
Ide code export and try command

toddtarsi:


toddtarsi:
Ok this will probably cause bugs and there is no try command yet but we have code export and a logging window so... worth it!

PR #1509, opened 2022-07-27
poorly copy a simplified version of electron-reloader

toddtarsi:


packages/selenium-ide/src/browser/components/AppPanelWrapper.tsx
```diff
@@ -15,6 +15,7 @@ const AppPanelWrapper: FC<AppPanelWrapperProps> = ({
     () => `fill ${horizontal ? 'outside-v-nav' : 'outside-nav'}`,
     [horizontal]
   )
+  console.log('Hey');
```
potapovDim, line 18: Should it be here ?

toddtarsi: @potapovDim - Ah! Hello there! Honestly, we kinda walked away from this a bit. One of the challenges is that the main process reload would have to pass all application state in, and without that, its sort of disruptive to your workflow because it just kills your process and drops you back on the splash page.

toddtarsi: And we just haven't prioritized passing in that much state via CLI flags.

toddtarsi: I'll close this PR for now and we can revisit this another time, but for now this is kinda a won't do thing. Thanks for looking over my PR!

Issue #1524, opened 2022-10-03
electron-chromedriver folder


giuliohome:
I followed the readme instruction on Ubuntu in Windwos WSL 2.

But I had to do the following

```
cp -r node_modules/electron-chromedriver/ packages/selenium-ide/node_modules/
```

(i.e. to copy the `electron-chromedriver` folder from the root's `node_modules` to the `node_modules` in the `packages/seleniumide`)

to make it work!

toddtarsi:
Hm! That's strange to me. What version of yarn / NodeJS are you building on? We specifically build using the 'nohoist' flag from yarn here.

https://github.com/SeleniumHQ/selenium-ide/blob/74fa23e5a7db0ea1b95917e7901cc59367a37e96/packages/selenium-ide/package.json#L92

This should keep these packages from getting lifted up to the parent directory, but I've messed many things up before haha.

giuliohome:
Hm, interesting. After a quick glance at the source, it seems you check the driver path here:

https://github.com/SeleniumHQ/selenium-ide/blob/3faf7291a91f5da242220ba3d8b64e54c790995e/packages/selenium-ide/src/main/session/controllers/Driver/start.ts#L67

which is resolving to

https://github.com/SeleniumHQ/selenium-ide/blob/3faf7291a91f5da242220ba3d8b64e54c790995e/packages/selenium-ide/src/main/session/controllers/Driver/start.ts#L36-L42

and I _suspect_ it simply points to the the `node_modules` in the `packages/selenium-ide` at runtime (irrespectively of nohoist in yarn workspaces... and btw [nohoist is deprecated in yarn3](https://yarnpkg.com/getting-started/migration#replace-nohoist-by-nmhoistinglimits)).

giuliohome:
My proposed fix

```bash
$ git diff packages/selenium-ide/
```

```diff
diff --git a/packages/selenium-ide/src/main/session/controllers/Driver/start.ts b/packages/selenium-ide/src/main/session/controllers/Driver/start.ts
index 85a96393..c86b7e81 100644
--- a/packages/selenium-ide/src/main/session/controllers/Driver/start.ts
+++ b/packages/selenium-ide/src/main/session/controllers/Driver/start.ts
@@ -35,7 +35,7 @@ const getDriver = ({ browser, version }: BrowserInfo) =>
     ? path.resolve(
         path.join(
           __dirname,
-          '..',
+          '..','..','..',
           'node_modules',
[...]

PR #1526, opened 2022-10-04
fix issue #1524

giuliohome:
Before submitting your PR, please check our [contributing](https://github.com/SeleniumHQ/selenium-ide/blob/master/CONTRIBUTING.md) guidelines.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description
The path for chrome driver is wrongly set when starting selenium ide. This is currently a blocking error.

### Motivation and Context
See my [comment there](https://github.com/SeleniumHQ/selenium-ide/issues/1524#issuecomment-1266477370)

### Types of changes

### Checklist

### DIFF
--- a/packages/selenium-ide/src/main/session/controllers/Driver/start.ts
+++ b/packages/selenium-ide/src/main/session/controllers/Driver/start.ts
@@ -35,7 +35,7 @@ const getDriver = ({ browser, version }: BrowserInfo) =>
     ? path.resolve(
         path.join(
           __dirname,
-          '..',
+          '..','..','..',
           'node_modules',
           'electron-chromedriver',
           'bin',
@@ -45,7 +45,7 @@ const getDriver = ({ browser, version }: BrowserInfo) =>
     : path.resolve(
         path.join(
           __dirname,
-          '..',
+          '..','..','..',
           'files',
           resolveDriverName({
             browser,


toddtarsi:
@giuliohome - I wrote a PR migrating from yarn v1 to pnpm this morning. I wanted to know if you'd consider this as an alternative to changing the pathing stuff on that chromedriver executable.

https://github.com/SeleniumHQ/selenium-ide/pull/1528

toddtarsi:
@giuliohome - I like that fix somewhat, but relying on hoisting outside of the package dependencies itself seems a bit hazardous. Pnpm is more up-to-date and consistent with keeping things local here and offers better controls for this. I also like dropping lerna because it has some strange dependencies these days, like Python 3.

giuliohome:
Ok, thank you, I'll try your suggestion, yes maybe pnpm could be good in my scenario, I remember I tried it some times ago and it was good.

toddtarsi:
@giuliohome - Would you try checking out the `switch-to-pnpm` branch and seeing if installing and building via pnpm works for you? I updated the readme as well, but the steps on that branch are this:

```
pnpm -r i; pnpm build; pnpm start;
```

giuliohome:
I'm doing this now... will let you know soon

giuliohome:
Ok, it works (I'd say yarn3 was faster maybe, but... ok).

giuliohome:
**Off Topic**
Excuse me, a question _unrelated_ to yarn3 and pnpm:
I can't export the code when I right click on a test: is this a separate issue to be opened or what?

toddtarsi:
@giuliohome - Yeah, yarn3 might be faster. I'm okay with doing it that way too. The only thing we'd have to do the other way is make sure that asarUnpack thing is updated and works by compiling with `build:electron`. If that isn't in the right place then it doesn't get bundled into the binary and the releases stop working. It was a whole headache trying to get that right.

toddtarsi:
@giuliohome - This is my fault. I am distributing an npm module for it right now:

```
https://npmjs.com/package/@seleniumhq/code-export
```

I have a PR where I was working on folding it into v4 of the IDE still:

https://github.com/SeleniumHQ/selenium-ide/pull/1507

Basically, everything in the IDE pipeline right now is configured to only use typed files and fail on a single type error. 
The problem is that the function shapes here are highly polymorphic and a really bad time to type, and also extremely redundant with the six different modules, so I haven't gotten that working 100% yet from IDE. I have a couple of the language bindings typed, but there's like four or five left to do and such.

giuliohome:
> @giuliohome - Yeah, yarn3 might be faster. I'm okay with doing it that way too. The only thing we'd have to do the other way is make sure that asarUnpack thing is updated and works by compiling with `build:electron`. If that isn't in the right place then it doesn't get bundled into the binary and the releases stop working. It was a whole headache trying to get that right.

I will re-try the whole build with this modification, all from the beginning with git clone (and this change), just to see if this problem is something I can check myself out...

toddtarsi:
@giuliohome - Sounds good. Thanks for calling this out and looking into this. Sorry to run you through this multiple times, I know this part is not fun. For the binary, you'll want to install the matching one for your OS from packages/selenium-ide/dist. It's a bit weird, I had to try to install a couple of them to find the one that worked for my Mac.

Once you find the one that installs right, you'll know it worked if you can load a project at all. Failing to load chromedriver has been wired up to crash things pretty aggressively so I find out from users more quickly.

giuliohome:
These are the commands I have tested again

```bash
rm -rdf selenium-ide/
git clone https://github.com/giuliohome/selenium-ide.git
cd selenium-ide/
git checkout patch-2
yarn set version berry
yarn
yarn build
yarn start
```

It seems all ok to me, but I don't want you accept something that you deem hazardous. Even though I can't see how it would "break the latest binary in alpha", it's not worth taking this risk, if there is this possibility. Feel free to close this PR, no worries. We can really consider this problem closed and resolved. Thank you again

toddtarsi:
Ok, I'll close this for now. Only because checking stuff on the binary is the opposite of a fun time. Thank you for doing these tests. If pnpm gives trouble, I'll keep your fix in mind and come back to go to yarn v3

Issue #1525, opened 2022-10-03
find element from label


giuliohome:
## 🚀 Find element for humans: by label

When I record a click on a button, I want to find it by looking for the label text. 
This is what a human being actually does.

## Motivation

Modern frameworks like React or Angular are based on components and state function models, thier styles don't provide ids to html elements or otherwise those ids are destryed and created again during effects and transitions.

## Example

This is an example of what I want to do in this enhancemet.
The point of intervention could be e.g. [here where](https://github.com/SeleniumHQ/selenium-ide/blob/trunk/packages/selenium-ide/src/browser/windows/PlaybackWindow/preload/record-handlers.ts#L119) the Playback window subscribes the click event handler to build the “low code” from the event target (seen as a html element) or more in-depth [where a locator](https://github.com/SeleniumHQ/selenium-ide/blob/74fa23e5a7db0ea1b95917e7901cc59367a37e96/packages/selenium-ide/src/browser/windows/PlaybackWindow/preload/locator-builders.ts#L64) is obtained from a builder.

For example, imagine a modern material design like this datepicker 

```python
from selenium import webdriver
driver = webdriver.Chrome()
driver.get('https://mui.com/x/react-date-pickers/getting-started/')
```

the process of finding the element to click (I'd say "find element for humans" https://twitter.com/giuliohome_2019/status/1576301123828187137 ... ) should be:

1.  find the label by looking at the label text
2. **find the id of the html element from the label `for` attribute!**
3. finally click on the element found by id

```python
def find_element_by_label(self, label_txt):
  found_label = self.driver.find_element(by = By.XPath, value = ".//label[text()='{label_txt}'])
  found_key = found_label.get_attribute('for') # my copyright
  return self.driver.find_element(by=By.ID, value = found_key)
```

the opposite during the **recording** process

1. the locator builder gets the id of the element from the event target of the event handler (this is the current as-is situation)
2. now check if there is a label for that element (i.e. a label having a `for` attribute with such an id, see the xpath code below)!
3. define a new builder locator that runs the above said `find_element_by_label`

the point 2 corresponds to something like the following

```python
reverse_for = driver.find_element('xpath',f'//label[@for={mykey}]')
```

A complete snippet of this eaxample's process, back and forward

```python
from selenium import webdriver
driver = webdriver.Chrome()
driver.get('https://mui.com/x/react-date-pickers/getting-started/')
[...]

PR #1530, opened 2022-10-05
feat #1525 text() in xpath:attributes locator

giuliohome:
### Description
Considering the html element's **textContent** in the locator builder `xpath:attributes`, 

### Motivation and Context
Please read the [thread here](https://github.com/SeleniumHQ/selenium-ide/issues/1525).

### Types of changes

### Checklist

### DIFF
--- a/packages/selenium-ide/src/browser/windows/PlaybackWindow/preload/locator-builders.ts
+++ b/packages/selenium-ide/src/browser/windows/PlaybackWindow/preload/locator-builders.ts
@@ -384,7 +384,6 @@ LocatorBuilders.add('xpath:img', function (this: LocatorBuilders, _e) {
 LocatorBuilders.add('xpath:attributes', function (this: LocatorBuilders, e) {
   const PREFERRED_ATTRIBUTES = [
     'id',
-    'textContent', //  it is not properly an "attribute"
     'name',
     'value',
     'type',
@@ -404,22 +403,22 @@ LocatorBuilders.add('xpath:attributes', function (this: LocatorBuilders, e) {
         locator += ' and '
       }
       let attName = attNames[i]
-      locator += (attName == 'textContent' ? 'text()' : '@' + attName) + '=' + this.attributeValue(attributes[attName])
+      locator += '@' + attName + '=' + this.attributeValue(attributes[attName])
+      if (e.textContent && attName == 'type') {
+        locator += ' and text()=' + this.attributeValue(e.textContent)
+      }
     }
     locator += ']'
     return this.preciseXPath(locator, e)
   }
 
-  if (e.attributes || e.textContent) {
+  if (e.attributes) {
     let atts = e.attributes
     let attsMap: Record<string, string> = {}
     for (i = 0; i < atts.length; i++) {
       let att = atts[i]
       attsMap[att.name] = att.value
     }
-    if (e.textContent) {
-      attsMap['textContent'] = e.textContent
-    }  
     let names = []
     // try preferred attributes
     for (i = 0; i < PREFERRED_ATTRIBUTES.length; i++) {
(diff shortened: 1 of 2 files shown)


packages/selenium-ide/src/browser/windows/PlaybackWindow/preload/locator-builders.ts
```diff
@@ -403,19 +404,22 @@ LocatorBuilders.add('xpath:attributes', function (this: LocatorBuilders, e) {
         locator += ' and '
       }
       let attName = attNames[i]
-      locator += '@' + attName + '=' + this.attributeValue(attributes[attName])
+      locator += (attName == 'textContent' ? 'text()' : '@' + attName) + '=' + this.attributeValue(attributes[attName])
```
toddtarsi: @giuliohome - This PR looks plenty good and clean. A request I have is this:

Can you move this to its own LocatorBuilder.add block? That way if an xpath has an attribute, thats an option as well as text content?

Another reason I'd really appreciate this, is at some point I'd like to support custom rankings for LocatorBuilder strategies, and I think it'd be really cool if you could tune this strategy above or below xpath:attributes as desired.

Honestly, you could just say 'no' or 'things came up and I'm too busy now' and I'd still merge this. A good contribution is a good contribution, my OCD aside.

Thanks for the clean code, the hard work, and the excellent communication!

giuliohome: Well, there is something that is related to your comment and I want to say here, even if it is not an answer (then I will also answer your question).
There is a problem (that I guess could be fixed in one line) in my change: it is slightly different from what I wrote: it should have outputted - for example - `@type` and text(), not only text().
Now, before separating it into a different builder (here starts the answer to your question) I wonder whether we want to see this one-line fix before. Also because if I separate it, the two builders will end up returning either a text() or an `@attribute`: that could be suboptimal in some edge cases maybe?
But eventually yep I'll rewrite it into a different builder. No problem at, aside the fact that I will do that later, because I have to start something else now. I think this would not be a problem for both of us, because there are other things to be developed for this v4, that can be addressed also in parallel. You mentioned the export to code (that is probably more important than this). And I noticed that "send keys" does not capture an initial control+a on the text field (and for a beginner that could be a blocker problem, and, aside that, also the priority of multiple choices is somehow hidden and a beginner can fail to notice the alternatives).
So finally I agree with you. I think first I will add a second very quick commit and then later (not sure exactly when atm) I will rewrite this as a separate builder.
We can keep this PR open till it is completed, I'd say: no need to rush and merge it.
Thank you very much for your help and support.

toddtarsi: @giuliohome - I've been committing 47 file commit chunks directly to trunk for the last eight months. This code is fine. I will hold off approving and merging until you toss in your other immediate commit and then we'll just merge and release it and do another PR later. For now, getting your hard work into trunk is probably for the best. Also, you've done great self-propelling through the code. I've done very little beyond misleading you to believe you needed to use yarn v3. All and all, I'm honored to get help from a coder of your calibre.

toddtarsi: @giuliohome - Honestly, I don't understand xpath evaluations nearly as well as you do so I totally defer to you on that stuff. I just like making big messes of typescript spaghetti 😄

giuliohome:
I've done my fix... please let me know what you think
Then we'll see how to proceed.
Thank you

BTW You're really too kind! Thanks so much but it's not deserved on my side

toddtarsi (approved):
LGTM 👍 Merging! Thanks for the excellent contribution!

PR #1662, opened 2023-06-25
new language opts for custom test/suite completions

toddtarsi:
@matewilk - How does something like this look for your purposes?

The idea here is that you can have generateTestCompletion or generateSuiteCompletion just like the declaration. Then, being able to be ahead of and behind it, you should be fine I think. 😄

### DIFF
--- a/packages/side-code-export/src/code-export/emit.ts
+++ b/packages/side-code-export/src/code-export/emit.ts
@@ -42,7 +42,10 @@ import { LanguageHooks } from './hook'
 export interface EmitterContext extends Omit<LanguageEmitterOpts, 'hooks'> {
   testLevel?: number
   commandLevel?: number
-  testDeclaration: string
+  suiteCompletion?: string
+  suiteDeclaration?: string
+  testCompletion?: string
+  testDeclaration?: string
   enableOriginTracing: boolean
   enableDescriptionAsComment: boolean
   hooks: LanguageHooks
@@ -351,6 +354,7 @@ async function emitTest(
   const {
     testLevel = 1,
     commandLevel = 2,
+    testCompletion,
     testDeclaration,
     terminatingKeyword,
     commandPrefixPadding,
@@ -411,7 +415,7 @@ async function emitTest(
   )
 
   // prepare result object
-  result.testDeclaration = render(testDeclaration, {
+  result.testDeclaration = render(testDeclaration!, {
     startingLevel: testLevel,
   }) as string
   result.inEachBegin = render(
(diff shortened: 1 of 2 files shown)


matewilk:
Hey @toddtarsi,

Thanks, that'll defo do the job, thanks for the PR!

matewilk (approved):
Looks amazing, gives the proper granularity over test/suite completion. Thanks!

toddtarsi:
@matewilk - I'll increment now

matewilk:
Thanks @toddtarsi!

matewilk:
Hey @toddtarsi ,

So I think something's missing here.
I thought I could use `generateTestCompletion` or `generateSuiteCompletion` as a function passed to `LanguageEmitterOpts` something like this:

```ts
export const opts: LanguageEmitterOpts = {
  // ......
  generateTestCompletion: function generateTestCompletion(name: string) {
    return `// test completion fn`;
  },
  generateSuiteCompletion: function generateSuiteCompletion(name: string) {
    return `// suite completion fn`;
  },
};
```

But I'm getting the following error:

> Waiting for the debugger to disconnect...
/Users/mwilk/Workspace/New Relic Synthetics Exporter/code-export-js-newrelic-synthetics/node_modules/side-code-export/dist/code-export/render.js:25
    if (typeof input === 'string' || 'statement' in input) {
                                                 ^

TypeError: Cannot use 'in' operator to search for 'statement' in undefined

Correct me if I'm wrong but shouldn't those two functions be assigned to a value and passed to `emit.test` / `emit.suite` in the `defaults.ts` like their `generate*` functions equivalent [here](https://github.com/SeleniumHQ/selenium-ide/blob/trunk/packages/side-code-export/src/code-export/defaults.ts#L53)?

I can still pass `testCimpletion` and `suiteCompletion` to `LanguageEmitterOpts` as a string like so:

```ts
export const opts: LanguageEmitterOpts = {
  // ...
  terminatingKeyword: "// terminating keyword",
  testCompletion: "// test completion str",
  suiteCompletion: "// suite completion str",
  // ....
};
```

But then I'm getting inconsistent results at the end of my generated script like so:

```ts
  // terminating keyword
// suite completion str
```

toddtarsi:
@matewilk - I completely agree. I wrote this for the generate* cases and thought the testCompletion things were just for intermediates to load onto emitter context else where. 

For my edification, which commands do you use from side-code-export to assemble your output stuff? There's a few ways to do it with this thing, and I want to make sure it's more of a bug and less of a code path thing.

toddtarsi:
@matewilk - Oh holy cow, checked your code reference! Yes, I totally missed that 😭 my bad!

If you PR a fix, I'd be happy to merge. Otherwise, I can do tonight.

matewilk:
If you could take it on that would be awesome, I'm swamped with preparing a presentation and other resources about this exact project for this Friday and it would take me some time before I figured out what exactly need to change.

toddtarsi:
@matewilk - No issue. I'll have it tonight. Thank you for the deep dive and looking at the code. Your help has been invaluable.

Issue #1673, opened 2023-07-11
I have a problem to get textCode.body from exportTestToFormat method of OutputFormatsController


mKabouri:
I want to get textCode.body  from exportTestToFormat method of OutputFormatsController. I created a new file:
```
var codeExport: string;

export function setCodeExport(newValue: string) {
    codeExport = newValue;
}

export function getCodeExport() {
    return codeExport;
}
```
I add setCodeExport(textCode.body) before return in exportTestToFormat method. After exporting a test in a language in selenium ide. I try to print with console.log getCodeExport() in another file. I always got undefined.

toddtarsi:
@mKabouri - Oh shoot, I see that export broke a while back. It appears to be infinitely looping. Both my kids are sick and one had to go to the hospital yesterday, so I'll try and find time to fix tonight, but it might have to wait for the weekend. Sorry about that!

toddtarsi:
@mKabouri - Please, give it a shot again on the latest version at your earliest convenience. Code-export should be repaired now. I accidentally had a case where it would compile forever until stack overflow, should be resolved now 😄

mKabouri:
I still have the same problem. I can explain what I want to do more clearly. I have created a new component inside the browser folder, and I created a new file that contains the code above. In the other hand, I add in exportTestToFormat method just before return this line of code:

```
        setTextareaContent(testCode.body);
```

The idea is to display the test code in my component when I export the test in a given language.

toddtarsi:
@mKabouri - Can you PR and I'll review from there? I am much more helpful looking at code than discussing it

mKabouri:
@toddtarsi I have PR. Thank you for your time.

toddtarsi:
@mKabouri - I understand now. Thank you for providing a code example. You want a code-export preview.

toddtarsi:
@mKabouri - Okay, having looked, I think I explained your issue. Let me know if you'd like any clarification, and nice feature!

PR #1675, opened 2023-07-19
PR for #1673 issue: Trying display test code in a text area

mKabouri:
This is a PR for #1673 issue. Thank you @toddtarsi.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test:e2e": "jest --testMatch \"**/packages/**/__tests__/**/*.e2e.js\"",
     "lint": "pnpm run lint:scripts",
     "lint:scripts": "eslint --ignore-pattern node_modules --ignore-pattern third-party --ignore-pattern dist --ignore-pattern build --ignore-pattern json --ext .ts,.tsx --ext .js packages/",
-    "test:ide": "npm-run-bg -s 'http-server -p 8080 ./tests/static::Available on::8080' 'node ./packages/side-runner/dist/bin.js -t 15000 ./tests/examples/*.side'",
+    "test:ide": "npm-run-bg -s 'http-server -p 8080 ./tests/static::Available on::8080' 'node ./scripts/ide-runner.js -t 15000 ./tests/examples/*.side'",
     "test:side-runner": "npm-run-bg -s 'http-server -p 8080 ./tests/static::Available on::8080' 'node ./packages/side-runner/dist/bin.js -t 15000 ./tests/examples/*.side'",
     "test:side-runner:ci": "npm-run-bg -s 'http-server -p 8080 ./tests/static::Available on::8080' 'node ./packages/side-runner/dist/bin.js -c \"goog:chromeOptions.args=[headless,no-sandbox] browserName=chrome\" -t 15000 ./tests/examples/*.side'",
     "typecheck": "tsc --noEmit --composite false",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -411,7 +411,7 @@ importers:
         specifier: ^4.0.0-alpha.4
         version: link:../side-model
       '@seleniumhq/side-runtime':
-        specifier: ^4.0.0-alpha.27
+        specifier: ^4.0.0-alpha.28
         version: link:../side-runtime
       lodash:
         specifier: ^4.17.21
@@ -473,11 +473,8 @@ importers:
         version: link:../side-code-export
     devDependencies:
       '@seleniumhq/side-api':
-        specifier: ^4.0.0-alpha.26
+        specifier: ^4.0.0-alpha.28
         version: link:../side-api
-      '@seleniumhq/side-runtime':
-        specifier: ^4.0.0-alpha.26
-        version: link:../side-runtime
 
   packages/side-migrate:
     dependencies:
@@ -509,7 +506,7 @@ importers:
         specifier: ^4.0.0-alpha.4
         version: link:../side-model
       '@seleniumhq/side-runtime':
-        specifier: ^4.0.0-alpha.27
+        specifier: ^4.0.0-alpha.28
         version: link:../side-runtime
       commander:
         specifier: ^11.0.0
@@ -3163,8 +3160,8 @@ packages:
     resolution: {integrity: sha512-+TLf8itRJHRI4dlTErDFtLeWpjDRz7M1VudMOX0iGyQ1PCEpmYuS2w3qfaeU7zRpUBhHyMlLrk+Ge5I6Fah65Q==}
     dev: false
 
-  /@seleniumhq/side-runtime@4.0.0-alpha.26:
-    resolution: {integrity: sha512-tD8WPka0QaXtHGZ3OEP1Y66P6WSfB//CJEsLdTVnKfkXYWy9GUrPp1KyIqw8nxDfkga6LbWd5HFgyiicuXX/cw==}
+  /@seleniumhq/side-runtime@4.0.0-alpha.27:
+    resolution: {integrity: sha512-t6rDtq+jAF/g0iXrX5Bv2z9Q41EpBF9o8PJCOEe2qYof+LOk5fJf8FSyHEATW2Bze44VLWrkljUCPbYdryeTGg==}
     dependencies:
       '@seleniumhq/side-commons': 4.0.0-alpha.1
       '@seleniumhq/side-model': 4.0.0-alpha.4
@@ -10236,7 +10233,7 @@ packages:
     hasBin: true
     dependencies:
       '@seleniumhq/side-model': 4.0.0-alpha.4
-      '@seleniumhq/side-runtime': 4.0.0-alpha.26
+      '@seleniumhq/side-runtime': 4.0.0-alpha.27
       commander: 9.4.1
     transitivePeerDependencies:
       - bufferutil


toddtarsi:
@mKabouri - I promise I haven't forgotten about this. This weekend I'll take a look. Btw, v42 or latest is a lot more stable to develop on, so if you come up, you'll have a less bad time now

toddtarsi:
Might have to still build, I still need to figure out where electron-chromedriver and pnpm are putting things in binary

toddtarsi:
@mKabouri - Am I correct that you're trying to improve tooling for format authors?

packages/selenium-ide/src/main/session/controllers/OutputFormats/index.tsx
```diff
@@ -75,6 +76,8 @@ export default class OutputFormatsController extends BaseController {
     const filepath = outputPath.filePath as string
     const testCode = await fileWriter.emitTest(format, project, testName)
 
+    setTextareaContent(testCode.body)
```
toddtarsi: @mKabouri - This wouldn't work because you're trying to share variables across two processes. The main electron process can't set a singleton variable in the sandboxed preload browser process. You'll need an API call to get this.

toddtarsi:
I love the intent behind this. However, this is what you'll need to do to get what you want:

1. In OutputFormatsController, make a new async method, something like `getTestCodeFromIDAndFormat`. Just gut the bits you need for this. If you want to be super clean, you can make the old function use it, but no pressure. This will be what does the work in the backend to get the info and return it.
2. Repeat this same function signature in side-api. It would be in a file like `packages/side-api/commands/outputFormats/getTestCodeFromIDAndFormat.ts`. Add this to side-api so that you can call `await window.sideAPI.outputFormats.getTestCodeFromIDAndFormat(test.id, format.name)` or whatever. Follow a file pattern like `commands/projects/getActive.ts`, and also make sure that you make index.ts files in side-api to hoist everything here similarly.
3. Get your code the easy way using window.sideAPI in the component layers as needed. At this point, intellisense should be your friend.

toddtarsi:
@mKabouri - I should explain, the side-api package is a weird and kinda ugly abstraction, but I wanted a utility package for plugin authors so they would have the entire shape of the api without having to install electron as a dev package.

mKabouri:
@toddtarsi I tried this but the Project editor just turning. Is there another alternative, for example using ipcRenderer and ipcMain ?

toddtarsi:
@mKabouri - Doing it this way is worth it. Commit what you have and I'll help.

toddtarsi:
@mKabouri - I'll take a look in about 6 hours and commit or PR some of the work to your branch, depending on the access I have.

toddtarsi:
@mKabouri - I'm sorry, I'm a bit too tired tonight. I'm going to try to look at this tomorrow. You're on the right track with the side-api additions!

packages/selenium-ide/src/browser/windows/ProjectEditor/components/ExportTest/ExportTest.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { getTextareaContent, setTextareaContent } from "./utils";
+import "./ExportTest.css"
+import { test } from "@seleniumhq/side-api";
+import { format } from "path";
+
+const ExportTest = () => {
+    const [body, setBody] = useState(getTextareaContent());
+
+    useEffect(() => {
+        window.sideAPI.outputFormats.getTestCodeFromIDAndFormat(test.id, format.name).then((data) => setTextareaContent(data));
```
toddtarsi: @mKabouri - Does this work now? This code looks immensely better

mKabouri: I think that I have a problem in getting format.name

mKabouri: It doesn't work. I get this error:

11:11:04.853 > [PROCESS] Unhandled Promise Rejection
11:11:04.854 > - - - - - - - - - - - - - - - - - - -
11:11:04.855 > TypeError: Cannot read properties of undefined (reading 'name')
    at OutputFormatsController.getTestCodeFromIDAndFormat (C:\Users\kabou\Desktop\stage2A\ide\mugnsoft-electron\packages\selenium-ide\build\webpack:\src\main\session\controllers\OutputFormats\index.tsx:92:65)
    at doAPI (C:\Users\kabou\Desktop\stage2A\ide\mugnsoft-electron\packages\selenium-ide\build\webpack:\src\main\api\classes\Handler.ts:54:22) 
    at IpcMainImpl.<anonymous> (C:\Users\kabou\Desktop\stage2A\ide\mugnsoft-electron\packages\selenium-ide\build\webpack:\src\main\api\classes\Handler.ts:65:22)
11:11:04.857 > - -

mKabouri: It doesn't work. I get this error:

11:11:04.853 > [PROCESS] Unhandled Promise Rejection
11:11:04.854 > - - - - - - - - - - - - - - - - - - -
11:11:04.855 > TypeError: Cannot read properties of undefined (reading 'name')
    at OutputFormatsController.getTestCodeFromIDAndFormat (C:\Users\kabou\Desktop\ide\selenium\packages\selenium-ide\build\webpack:\src\main\session\controllers\OutputFormats\index.tsx:92:65)
    at doAPI (C:\Users\kabou\Desktop\ide\selenium\packages\selenium-ide\build\webpack:\src\main\api\classes\Handler.ts:54:22) 
    at IpcMainImpl.<anonymous> (C:\Users\kabou\Desktop\ide\selenium\packages\packages\selenium-ide\build\webpack:\src\main\api\classes\Handler.ts:65:22)
11:11:04.857 > - -

This what I do:
```
window.sideAPI.outputFormats.getTestCodeFromIDAndFormat('csharp-nunit', test.id)
.then((data) => setTextareaContent(data));
```

toddtarsi: @mKabouri - That seems like it should work. Yeah, importing format from a random node module definitely wouldn't work.

packages/selenium-ide/src/browser/windows/ProjectEditor/components/ExportTest/utils.ts
```diff
@@ -0,0 +1,12 @@
+var textareaContent: string
+
+export function setTextareaContent(newValue: string) {
```
toddtarsi: I would drop this entire file. It is misleading

packages/selenium-ide/src/browser/windows/ProjectEditor/components/ExportTest/ExportTest.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from "react"
+import { getTextareaContent, setTextareaContent } from "./utils";
+import "./ExportTest.css"
+import { test } from "@seleniumhq/side-api";
+import { format } from "path";
```
toddtarsi: Right, this module ahouldn't even work in the browser. Gotta drop this line

mKabouri: Yes, I have removed it and specifying a language like 'csharp-nunit' instead.

toddtarsi: Well that sounds rad. Any luck?

mKabouri: No, getTextareaContent always returns 'None'. I don't know if we have to call getTestCodeFromIDAndFormat() method in another file of main ?

toddtarsi: This is usually how you do this in React:

```
const [code, setCode] = React.useState('');
React.useEffect(() => window.sideAPI.xyz(whatever, csharp).then(setCode), [csharp]);
```

If you're not doing something like this, it probably won't work

toddtarsi:
@mKabouri - I think you need to supply some state for export format as well. You might be better off coming at this from a menu / hotkey angle so that format iteration and selection is solved (example here: https://github.com/SeleniumHQ/selenium-ide/blob/acf86928d1c9950b9c60d320cf0c1bcbb996a736/packages/selenium-ide/src/main/session/controllers/Menu/menus/suiteManager.ts).

This might sound shady, but I'm starting to work on a paid plugin that will basically "supercharge" (buzzword I'm sorry) export format development and playback as well, so while I'm supportive of any effort to democratize and add good code-export features and tooling, I'm planning to solve some of these DX problems to a pretty extreme degree via a cheap plugin.

toddtarsi:
Oh, I tried it out last night! But yeah, it seems a little too incomplete right now.

packages/selenium-ide/src/browser/windows/ProjectEditor/components/ExportTest/ExportTest.css
```diff
@@ -0,0 +1,7 @@
+.test-export {
```
toddtarsi: Unfortunately, I don't think theres really a pipeline to use css this way, so I'd just recommend creating these as inline styles in the React component and using them that way. I appreciate you hewing close to the correct way on this stuff but I'm okay with bad perf for now.

mKabouri:
I console.log test.id in the browser side it always prints '-1' in the console, and test is imported from side-api. I don't know if this is the problem because when I create a new test it has also an id of '-1'.

toddtarsi:
@mKabouri - Can I see a reference to a line of code for what you're experiencing?

toddtarsi:
@mKabouri - Said differently, can you show me how you're always getting -1?

mKabouri:
In the line before window.sideAPI.get....("language", test.id). I just console log `console.log("Debug: ",  test.id);`. I got 'Debug: -1'

packages/selenium-ide/src/browser/windows/ProjectEditor/components/ExportTest/ExportTest.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from "react"
+import { getTextareaContent, setTextareaContent } from "./utils";
+import "./ExportTest.css"
+import { test } from "@seleniumhq/side-api";
```
toddtarsi: @mKabouri - Remember, this is in an isolated process, separate from the backend. Importing variables won't work just like when you tried setTextAreaContent from the backend and couldn't use it in the browser.

toddtarsi: Instead, you'll need to either pass in the test ID from further up ala react props, or you'll need to use the sideAPI to retrieve this:

```
const {project, state} = await window.sideAPI.state.get()
const testID = state.activeTestID
```

As always, intellisense should be your guiding star here

mKabouri: @toddtarsi  And it works hhhh. Thank you for your time. I really appreciate. It will be a good idea to add a README in sideAPI.

toddtarsi: @mKabouri - Absolutely. My plan is to expose the entirety of the side api directly to plugins, so I'll definitely be working on that in the next weekend or two

PR #1719, opened 2023-10-10
ability to compose selectors, api layer

toddtarsi:
This __should__ expose the capability in the recorder to reorder selectors and persist this across sessions as part of state.

This still needs a consuming component set in the project editor (in system config), that is a drag and drop list with these selectors.


toddtarsi:
Folding this into trunk once tests pass.

